PR #6, opened 2023-05-10
Draft version of basic superbol-mode for GNU/Emacs

nberth:


README.md
```diff
@@ -65,6 +65,29 @@ In the `superbol` field past the path to the `superbol` executable.
 
 You can check the documentation on using the extension on [this page](https://ocamlpro.github.io/superbol-vscode-platform/sphinx).
 
+### GNU/Emacs mode
```
lefessan, line 68: Just for the record : we should avoid adding this kind of technical documentation in the README.md file. Technical documentation should go to the Sphinx doc, while the README.md is just an entry point describing the project and its features, and linking to the Sphinx documentation for more technical details.

GitMensch:
Looks interesting. I volunteer to test a version for Vim ;-)

lefessan:
> Looks interesting. I volunteer to test a version for Vim ;-)

@emilienlemaire You told me that using it in `vim` would require uploading it to some online repo, do you have any other recipe ?

emilienlemaire:
I was talking about Neovim (which has built-in lsp support), I'm not sur how it works on plain vim (which don't have built in lsp support), I can check for it, but this probably would imply a dependency on another plugin

README.md
GitMensch: Actually, as it is now, I find it quite helpful for the initial setup.

nberth: Now, I think the fact that we can debate the place for the documentation means the GNU/Emacs mode itself does not fit in this particular repository very well. Wouldn't it be better to have a (however small) separate standalone `superbol-mode` repository with just the `README.md` (as with many other modes for Emacs) ?

GitMensch: A minimal repo with the emacs folder and the README.md seems most useful.
The "common" parts of the LSP could be a git external set to a fixed revision of this repo that was tested against emacs.

PR #7, opened 2023-06-06
Documentation

emilienlemaire:


emilienlemaire:
This adds some documentation, I'll add more on superbol directly

PR #9, opened 2023-07-13
Fix package name in build scripts

nberth:


emilienlemaire (approved):
All good, I thought I already changed those 😅
Ok to merge after CI passed 👍

nberth:
Actually what is the CI doing? Is it supposed to reach the `make compile` stage?

emilienlemaire:
As of now it just does a `drom build`, but that  has been useful to check if everything compiled well on windows

nberth:
Do you think it's relevant to complete the build process in the CI (with yarn and all)?

emilienlemaire:
It would be for deployment maybe, to generate the `.vsix` file automatically, but for now it will just slow down the CI which is already not that fast

nberth:
Ok, for a deployment test maybe. That was just to check what the CI is used for.

PR #12, opened 2023-09-15
Remove fold on ephemerons

nberth:
Before this change we had a warning about a function not being available in OCaml 5.


lefessan:
There are too many unrelated commits in this PR. If you want a review, I think it needs to be splitted in multiple simpler PRs. Typically, changes to `drom` files that modify many generated files should be done in a specific PR, with a single commit containing all the changes to generated files.

PR #13, opened 2023-09-15
LSP improvements

nberth:
This improves the LSP via:
- A refactoring of its internal code, and add handling for every kind of Jsonrpc messages that was previously missing;
- The emission of semantic tokens ("semtoks") for comments and (some) pre-processor directives;
- A new ability to reply with ranged semtoks (*i.e.* on a restricted portion of a file);
- New support for folding-range requests.

Slight alterations of the AST representation for `PERFORM` and `SEARCH (ALL)` statements also come along.

At last, included changes remove the completion capability, that is only partially implemented.


src/lsp/cobol_indent/indenter.ml
```diff
@@ -62,8 +62,8 @@ let indenter ~source_format (str:string) (rdl:indent_record list) range =
   String.concat "\n" strl
 
 (*indent a range of file, with the default indent_config*)
-let indent_range' ~source_format ~range ~file =
-  let file_content = Ez_file.V1.EzFile.read_file file in
+let indent_range' ~source_format ~range input =
```
lefessan, line 65: I don't understand this change: you are replacing a name that is very specific `file` with a generic name `input`, where you completely lose the fact that it MUST be a file.

nberth: Indeed, that was remains of some quick attempt to relax the limitation mentioned in the comment below. Now that's been reverted.

PR #14, opened 2023-09-18
Actually copy `superbol-free` executable after build

nberth:


nberth:
@lefessan I'm not sure this is how we should handle the resulting executable files. Maybe `drom` has better ways, so I won't merge that PR right now.

PR #17, opened 2023-09-22
Rewindable parser

nberth:
These changes add the ability to restart parsing only a few tokens before any document change. This should greatly improve performance, especially when editing the middle or end of large files.

Changes also include some refactoring of the parser and pre-processor, with a unification of their interface (notably, with options always passed using a single record).

Some tests of the rewinding feature come as well, along with fixes for some bugs they helped discover ;-).


src/lsp/cobol_ast/misc_descr.ml
```diff
@@ -14,25 +14,26 @@
...
-    date_written: string with_loc option;
-    date_compiled: string with_loc option;
-    security: string with_loc option;
-  }
-[@@deriving show, ord]
+  informational_paragraph with_loc list
+[@@deriving ord]
+
+and informational_paragraph =
+  informational_paragraph_header * comment_entry with_loc
```
lefessan, line 25: What is the rational behind this change ? It's much easier to lookup the information in the record version than in the associated list, no ?

Maybe what we are struggling with is the fact that we chose to have only one AST to do everything, when we should probably have a "summarized" version of the AST: one AST that is close to the source, and one that only contains the interesting information for analyzis and compilation.

src/lsp/cobol_ast/misc_descr.ml
```diff
@@ -655,7 +656,8 @@ let pp_select_clause ppf = function
   | SelectRecordKey { key; source } ->
     Fmt.pf ppf "RECORD %a%a"
       pp_qualname key
-      Fmt.(elist ~prefix:(any "SOURCE ") ~sep:sp pp_name') source
+      Pretty.(list ~fopen:"SOURCE " ~fsep:"@ " ~fclose:"" ~fempty:""
```
lefessan, line 659: Why `Pretty.(...)` instead of `(Pretty.list ...)` ?

src/lsp/cobol_ast/misc_descr.ml
```diff
@@ -721,21 +725,46 @@ let pp_arithmetic_mode ppf = function
...
-    ++ any "."
-  )
+  Fmt.(any "OPTIONS.@ " ++
+       box (list ~sep:sp (pp_with_loc pp_options_clause)) ++
+       any ".")
+
+
+(* --- *)
+
+let pp_informational_paragraph_header ppf = function
```
lefessan, line 749: Functions with arity 2 should be defined with 2 arguments, `let f x = function ...` is bad for readability.

src/lsp/cobol_ast/misc_descr.ml
```diff
@@ -721,21 +725,46 @@ let pp_arithmetic_mode ppf = function
...
+  | DateWritten -> Fmt.pf ppf "DATE-WRITTEN"
+  | Installation -> Fmt.pf ppf "INSTALLATION"
+  | Remarks -> Fmt.pf ppf "REMARKS"
+  | Security -> Fmt.pf ppf "SECURITY"
+
+let pp_comment_entry: comment_entry Pretty.printer =
+  Fmt.(list ~sep:sp string)
+
+let pp_informational_paragraph: informational_paragraph Pretty.printer =
+  Fmt.(any "@[<4>" ++                       (* <- indent by 4 to avoid Area A *)
```
lefessan, line 762: Weird that there is not a construct to do that, like symbolically, ie create a box with an indentation with automatic close at the end. All these strings as format descriptions are heaven for bugs in formatting.

src/lsp/cobol_parser/cobol_parser.ml
```diff
@@ -18,33 +18,18 @@ module PTree = PTree
 module PTree_visitor = PTree_visitor
```
lefessan, line 18: Not in the patch, but why does `cobol_parser` exports a visitor interface ? For me, there should be a library `cobol_visitors`, and `cobol_ast` and `cobol_parser` should be independent of it, i.e. it should be possible to use the parser without using the visitors.

src/lsp/cobol_parser/parser_outputs.ml
```diff
@@ -0,0 +1,32 @@
...
+(*                                                                        *)
+(* All rights reserved.                                                   *)
+(* This source code is licensed under the GNU Affero General Public       *)
+(* License version 3 found in the LICENSE.md file in the root directory   *)
+(* of this source tree.                                                   *)
+(*                                                                        *)
+(**************************************************************************)
+
+(** Gathers some types used to define outputs for the parser engine. *)
+(* We are only defining types here so an MLI would only be redundant. *)
```
lefessan, line 15: That's some kind of `types.ml` module, then ?

src/lsp/cobol_parser/parser_outputs.ml
```diff
@@ -0,0 +1,32 @@
...
+type tokens_with_locs = Grammar_tokens.token with_loc list
+
+type artifacts =
+  {
+    tokens: tokens_with_locs Lazy.t;
+    pplog: Cobol_preproc.log;
+    comments: Cobol_preproc.comments;
+  }
+
+type ('a, 'm) output =
```
lefessan, line 28: * I cannot see the use of the `'m` parameter. 
* I don't understand what `'a` and `'m` stand for, maybe using longer names that 1-character letters would help
* I don't see the point of using yet more GADTs. I think we should go the other way around: progressively getting rid of GADTs, because they make the code impossible to understand for anybody else than their author.

lefessan (commented):
I stopped when I found yet another GADT. I think COBOL is complex enough without adding OCaml complexity on top of it. Please, progressively remove GADTs instead of adding new ones...

src/lsp/cobol_ast/misc_descr.ml
bclement-ocp: What you describe is `Fmt.box ~indent:4`

src/lsp/cobol_ast/misc_descr.ml
nberth: probably so that `pp_name'` (below) indents properly.

src/lsp/cobol_ast/misc_descr.ml
nberth: I'd say the form with `function` reads much better for many specific cases, like pretty-printing procedures for instance (just need to visually match `pp_ … ppf` to spot those.

src/lsp/cobol_ast/misc_descr.ml
nberth: Ah yes, good idea ;-)

src/lsp/cobol_parser/cobol_parser.ml
nberth: Yes that's a change planned for a little later.

src/lsp/cobol_parser/parser_outputs.ml
nberth: Yes, but with a better name.

src/lsp/cobol_parser/parser_outputs.ml
nberth: Ok I'll document that aspect.

src/lsp/cobol_lsp/lsp_document.ml
```diff
@@ -99,28 +127,51 @@ let lazy_references ast cus defs =
...
+    end ptree
   in
-  { doc with artifacts; diags; parsed }
+  { doc with artifacts; rewinder; diags; parsed }
+
+let parse_and_analyze ({ copybook; _ } as doc) =
+  if copybook then                                                    (* skip *)
+    { doc with artifacts = no_artifacts; rewinder = None; parsed = None }
+  else
+    (* extract_parsed_infos doc @@ simple_parse doc *)
```
bclement-ocp, line 161:
```suggestion
```

ddeclerck (approved):

PR #20, opened 2023-09-26
Fixes to re-enable CI and cross-compilation

lefessan:


nberth:
@lefessan Please squash before merging to avoid keeping those useless commit messages around ;-)

PR #23, opened 2023-09-27
Keep abbreviated combined relation conditions as is in the parse tree

nberth:
These changes alter the AST representation for conditional expressions, where abbreviated forms are kept as is, instead of being re-written using standard binary relation operators and logical combinators.


src/lsp/cobol_ast/terms.ml
```diff
@@ -615,13 +630,20 @@ module FMT = struct
...
     | Expr e ->
         fmt "%a%a" ppf not_ pos pp_expression e
-    | Relation (a, o, b) ->
-        fmt "@[<1>%a(%a@ %s@ %a)@]" ppf
-          not_ pos pp_expression a ([%derive.show: relop] o) pp_expression b
+    | Relation rel ->
+        fmt "%a@[<1>(%a)@]" ppf not_ pos pp_binary_relation rel
+    | Abbrev (neg, rel, o, comb) ->
+        fmt "%a@[<1>(%a%a@ %a@ %a)@]" ppf
+          not_ pos not_ neg pp_binary_relation rel pp_logop o
```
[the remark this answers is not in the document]
nberth: Ah yes indeed, `not_` accepts a polarity. Thanks for testing!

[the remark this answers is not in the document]
bclement-ocp: Well that one was easily caught by the pretty-printer :)

src/lsp/cobol_ast/terms.ml
```diff
@@ -149,12 +149,27 @@ and expression =
...
   | ClassCond: expression * class_ -> [>simple_] cond (* exp = ident *)
   | SignCond: expression * signz -> [>simple_] cond (* exp = arith exp *)
   | Omitted: expression -> [>simple_] cond (* exp = ident *)
   | Not: _ cond -> [>complex_] cond
-  | Logop: _ cond * logop * _ cond -> [>complex_] cond (* TODO: move logop left *)
+  | Logop: _ cond * logop * _ cond -> [>complex_] cond
+
+and binary_relation =
+  expression * relop * expression
+
```
lefessan, line 162: Do you think it would be possible to add comments with the corresponding syntax in COBOL ? it's not always easy to match such types with what they encode, and reciprocally

src/lsp/cobol_ast/terms.ml
```diff
@@ -615,13 +630,20 @@ module FMT = struct
     | BXor -> "B-XOR"
   and pp_binop ppf o = string ppf (show_binop o)
 
+  and pp_binary_relation ppf (a, o, b) =
+    fmt "%a@ %s@ %a" ppf
+      pp_expression a ([%derive.show: relop] o) pp_expression b
```
lefessan, line 635: What is `[%derive.show: relop]` ? Why not use the `pp_relop` generated by `deriving show` ? it's not the same ?

nberth: Indeed, it's less opaque.

PR #26, opened 2023-09-27
Add an AST pretty-printer

bclement-ocp:
The pretty-printer formats code using a COBOL-like language that tries
to stay as compatible as possible with free-form COBOL. It is currently
not configurable (e.g. to output for different dialects).

Note that some issues are known regarding nested indexing (or indexing
within a function call), where either too much or not enough parentheses
may be added. This will be investigated and fixed in a separate PR.

The pretty-printer can be used with the `pp` command by passing the
`--parse` argument. To test the parser, the `--check` argument can be
passed to the `pp` command, which will check that re-parsing the
pretty-printed code gives back the same AST as the original.

The new `reparse` test checks the testsuite in the same way described
above.

Note: this includes both #23 and #24 that change the AST.

As a side note: the "check" comparison is sensitive to the way the comparison of AST terms is performed. It is very easy to miss cases in `compare_term`, which cause equal terms to be treated as different (because the `Stdlib.compare` fallback cares about position).


test/output-tests/reparse.ml
```diff
@@ -0,0 +1,76 @@
...
+(*                                                                        *)
+(**************************************************************************)
+
+open Format
+open Ez_file
+open FileString.OP
+open Cobol_preproc
+
+let deep_iter = FileString.(make_select iter_dir) ~deep:true
+let srcdir = try Unix.getenv "DUNE_SOURCEROOT" with Not_found -> "."
```
lefessan, line 20: Take a look at https://github.com/OCamlPro/superbol-studio-oss/blob/2b9c53e13855c3cfaaf60b9d0cfab34cc9ea9207/test/output-tests/preproc.ml#L19 to make the relocation of tests ok

bclement-ocp: I am not sure what the use for this is, can you clarify why/when tests are relocated?

bclement-ocp: Ah but this is upstream changes to the file I copied, OK

PR #28, opened 2023-09-28
Always print tests output relative to import/gnucobol/

lefessan:


test/output-tests/gnucobol.ml
```diff
@@ -40,12 +40,28 @@ let target =
...
+    in
+    Fmt.string ppf s
+*)
+
+let pp_relloc fmt s =
+  let path = EzString.split s '/' in
+  let rec iter path =
+    match path with
+    | [] -> s
+    | "import" :: "gnucobol" :: _ -> String.concat "/" path
```
nberth, line 60: While you're at it, why not strip the `import` altogether?

Also `ppf`  stands for pretty-printing-formatter (I think). `fmt` is the old-school name, that unfortunately can be mistaken for "format-string".

lefessan: Ok for `ppf`. For `import`, I don't want to match only on `gnucobol`, as it may appear above the source directory...

test/lsp/lsp_formatting.ml
```diff
@@ -401,10 +401,10 @@ let%expect_test "formatting-request-whole-program" =
   end_with_postproc [%expect.output];
   [%expect {|
     {"params":{"diagnostics":[],"uri":"file://__rootdir__/superbol.toml"},"method":"textDocument/publishDiagnostics","jsonrpc":"2.0"}
-    src/lsp/cobol_ast/raw_misc_sections_visitor.ml:66:
+    raw_misc_sections_visitor.ml:0:
```
nberth, line 404: @lefessan What was the problem of having precise source locations here?
To me the occasional need for an auto-promotion does not warrant the ugly hack used in this PR to remove the info.

nberth: These locations are very handy when implementing the partial visitors, even and in particular in the tests.

PR #29, opened 2023-09-29
Add an AST pretty-printer

bclement-ocp:
The pretty-printer formats code using a COBOL-like language that tries
to stay as compatible as possible with free-form COBOL. It is currently
not configurable (e.g. to output for different dialects).

Note that some issues are known regarding nested indexing (or indexing
within a function call), where either too much or not enough parentheses
may be added. This will be investigated and fixed in a separate PR.

The pretty-printer can be used with the pp command by passing the
--parse argument. To test the parser, the --check argument can be
passed to the pp command, which will check that re-parsing the
pretty-printed code gives back the same AST as the original.

The new reparse test checks the testsuite in the same way described
above.

Note: this includes https://github.com/OCamlPro/superbol-studio-oss/pull/24 that changes the AST.

As a side note: the "check" comparison is sensitive to the way the comparison of AST terms is performed. It is very easy to miss cases in compare_term, which cause equal terms to be treated as different (because the Stdlib.compare fallback cares about position).

Replaces #26


src/lsp/cobol_ast/abstract.ml
```diff
@@ -13,39 +13,54 @@
 
 module type MISC_SECTIONS = sig
   (* IDENTIFICATION_DIVISION *)
-  type informational_paragraphs [@@deriving show]
-  type options_paragraph        [@@deriving show]
-  type environment_division     [@@deriving show]
+  type[@deriving ord] informational_paragraphs
```
nberth, line 16: I'm not a huge fan of putting attributes/extensions here as I often search with regexps using strings like "`type <typename-prefix>`" when looking for type definitions sometimes (we all have our old-school ways of editing). I'm also wondering, wasn't `[@deriving show, ord]` enough to declare the `pp_` functions?

bclement-ocp: `[@@deriving show]` declares the `pp_` functions but also `show_` functions that are never used. I can move the deriving to paragraph attributes.

nberth: Well, we derive `show_` for most of the other types; we can keep those in favor of concision of the source code.

nberth:
BTW, this might deserve an additional rebase since I merged #24.

src/lsp/cobol_ast/abstract.ml
nberth: Ah but you did not implement the `show_` functions indeed.

bclement-ocp: I mean I can implement `show_` but figured that since it is not used it was not necessary :) I can change it to use `deriving show` and implement `show_` if you want.

bclement-ocp:
Should be properly rebased on current master now

src/lsp/cobol_ast/abstract.ml
nberth: Well… I guess that makes a lot of single-line definitions to add. I'd like to quickly merge some of my changes as well, but these touch the AST so we can ignore `show`'s for now.

PR #40, opened 2023-10-09
Add a set of `superbol switch` commands to manage GnuCOBOL installations

lefessan:
The current commands are:

* `superbol switch env [SWITCH] [--last]`: print shell commands to set environment variables (shortcut `superbol env`)
* `superbol switch import [--clear] [--no-set] [DIRS]`: import GnuCOBOL installations from directories DIRS
* `superbol switch list`: list known switches
* `superbol switch set [SWITCH] [--last]`: set the current default switch
* `superbol switch add [GNUDIR] [--switch SWITCH] [--no-set]`: add a specific installation directory as a switch
* `superbol switch build [DESTDIR] [--switch SWITCH] [--no-set] [--sudo]`: configure, build and install GnuCOBOL as a switch


lefessan:
FWIW, switches are numbered, starting from S01-xxx to SNN-xxx, in the order they are created. They can be selected using only any substring, i.e. `--switch S12` will select switch number 12.

PR #71, opened 2023-10-25
Indent text from buffer, not disk

bclement-ocp:
When indenting a file from the LSP, use the text from the current buffer, not the last version saved to disk (otherwise changes could disappear).

### DIFF
--- a/src/lsp/cobol_indent/indenter.ml
+++ b/src/lsp/cobol_indent/indenter.ml
@@ -62,30 +62,22 @@ let indenter ~source_format (str:string) (rdl:indent_record list) range =
   String.concat "\n" strl
 
 (*indent a range of file, with the default indent_config*)
-let indent_range' ~dialect ~source_format ~range ~file =
-  let file_content = Ez_file.V1.EzFile.read_file file in
-  (*
-    Not satisfied with the `Cobol_preproc.fold_text_lines`,
-    this function has an argument which is the name of file,
-    so when using lsp, every time using the formatting,
-    we must save the file before, it is not convenient.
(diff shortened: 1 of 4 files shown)


src/lsp/cobol_indent/indenter.ml
```diff
@@ -62,30 +62,22 @@ let indenter ~source_format (str:string) (rdl:indent_record list) range =
...
-let indent_range' ~dialect ~source_format ~range ~file =
-  let file_content = Ez_file.V1.EzFile.read_file file in
-  (*
-    Not satisfied with the `Cobol_preproc.fold_text_lines`,
-    this function has an argument which is the name of file,
-    so when using lsp, every time using the formatting,
-    we must save the file before, it is not convenient.
-     (* NB: not anymore. *)
-  *)
+let indent_range' ~dialect ~source_format ~range ~filename ~contents =
```
nberth, line 65: I'm not sure why we had those primes here. Could you take the opportunity of this small PR to get rid of them?

nberth (approved):
All fine with me. Just a minor suggestion.

src/lsp/cobol_indent/indenter.ml
bclement-ocp: This is because `indent_range` is exposed in `Cobol_indent` as the same function, but prints the result for CLI usage. I can make the printing happen in `command_indent_range.ml` instead of `cobol_indent.ml` but I'd rather do it in a separate PR.

nberth: Ok, let's delay that then.

PR #72, opened 2023-10-26
Fix form for customizing `superbol.path`

Stevendeo:
This PR fixes the `superbol.path` form that was not displayed because of a type error.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
           "description": "If something is selected, only format the selection"
         },
         "superbol.path": {
+          "type": "string",
           "default": "superbol-free",
           "description": "Path to the `superbol` command"
         }
--- a/src/lsp/superbol_free_lib/project.ml
+++ b/src/lsp/superbol_free_lib/project.ml
@@ -145,7 +145,9 @@ let contributes =
 
             Manifest.PROPERTY.string "superbol.path"
               ~default:"superbol-free"
-              ~description: "Path to the `superbol` command"
+              ~description:
+              "Name of the `superbol` command; path can be relative (deprecated) or \
+               absolute."
           ] )
     ~taskDefinitions: [
       Manifest.taskDefinition
--- a/src/lsp/superbol_free_lib/project.ml
+++ b/src/lsp/superbol_free_lib/project.ml
@@ -146,7 +146,7 @@ let contributes =
             Manifest.PROPERTY.string "superbol.path"
               ~default:"superbol-free"
               ~description:
-              "Name of the `superbol` excutable if available in PATH; \
+              "Name of the `superbol` executable if available in PATH; \
                otherwise may be an absolute path."
           ] )
     ~taskDefinitions: [
--- a/src/lsp/superbol_free_lib/project.ml
+++ b/src/lsp/superbol_free_lib/project.ml
@@ -146,8 +146,8 @@ let contributes =
             Manifest.PROPERTY.string "superbol.path"
               ~default:"superbol-free"
               ~description:
-              "Name of the `superbol` command; path can be relative (deprecated) or \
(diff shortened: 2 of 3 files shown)


nberth:
This form seemed to show correctly to me before. Was there a specific issue with its rendering?

While we are at it, is there a way to be more specific, and, for instance, tell it's a(n existing) directory that we'd want in this field?

Stevendeo:
> This form seemed to show correctly to me before. Was there a specific issue with its rendering?
On my side, it displays a link to the settings.json file to edit the option manually. Which version of vscode do you use? I am using 1.82.2.

> While we are at it, is there a way to be more specific, and, for instance, tell it's a(n existing) directory that we'd want in this field?
I'm trying to make variable substitution work, the doc says something like `${env:PATH}` should work, but it does not (at least on my version).

nberth:
> On my side, it displays a link to the settings.json file to edit the option manually. Which version of vscode do you use? I am using 1.82.2.

Hum not good news: I'm using 1.80.0. So I assume the current specs are a just becoming out-dated.

Stevendeo:
Variable substitution does not seem to work either on the last version (1.83.1), it would have been too easy ^^

Stevendeo:
Good thing, though: putting `superbol` alone in the field seems to be sufficient.

package.json
```diff
@@ -83,9 +84,11 @@
             "description": "The list of copybooks paths"
           },
           "sourceFormat": {
+            "type": "string",
```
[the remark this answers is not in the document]
nberth: Note this is generated from `src/lsp/superbol_free_lib/project.ml`. Some adjustments to task definition (that we don't seem to use yet — and a unification of variable names maybe) might be needed there.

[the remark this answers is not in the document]
Stevendeo: Check #78

src/lsp/superbol_free_lib/project.ml
```diff
@@ -145,7 +145,9 @@ let contributes =
 
             Manifest.PROPERTY.string "superbol.path"
               ~default:"superbol-free"
-              ~description: "Path to the `superbol` command"
+              ~description:
+              "Name of the `superbol` command; path can be relative (deprecated) or \
+               absolute."
```
nberth, line 150: This does not seem to convey the idea that one can put an executable name that will be search in `PATH`. Is it actually allowed?

GitMensch: Yes, it is allowed

GitMensch:
@nberth If you want changes to this PR then please review requesting changes. Otherwise I suggest to pull this in to get the type error fixed (we can easily adjust the description later).

src/lsp/superbol_free_lib/project.ml
nberth: Then I'd suggest something like "Name of the `superbol` executable if available in PATH; otherwise, may be an absolute path."

nberth (approved):
One small suggestion and we can pull this in ;-)

nberth (changes requested):
Wrong button on previous review: actually requesting a change.

src/lsp/superbol_free_lib/project.ml
```diff
@@ -145,7 +145,9 @@ let contributes =
 
             Manifest.PROPERTY.string "superbol.path"
               ~default:"superbol-free"
-              ~description: "Path to the `superbol` command"
+              ~description:
+              "Name of the `superbol` excutable if available in PATH; \
```
nberth, line 149: ex_**e**_cutable. (sorry that may be on me).

nberth (approved):
Ok to merge after fixing typo ;-)

PR #74, opened 2023-10-26
 Only enable indenter on free format 

### LINKED ISSUE #52: Disable auto-indent in fixed format
For alpha release, auto-indent should be disabled on fixed format, and only in the next release shall it be implemented (#45)

bclement-ocp:
The indenter also now uses inferred formats properly.

Fixes #52

Note: Currently includes #73; do not merge before #73.

### DIFF
--- a/src/lsp/cobol_indent/indenter.ml
+++ b/src/lsp/cobol_indent/indenter.ml
@@ -73,14 +73,10 @@ let indenter ~source_format (str:string) (rdl:indent_record list) range =
 
 (*indent a range of file, with the default indent_config*)
 let indent_range ~dialect ~source_format ~range ~filename ~contents =
-  let src_format =
-    (* Note: this value doesn't actually matter, it will be overriden
-       immediately by [fold_source_lines] calling [on_initial_source_format]
-       below. *)
-    match source_format with
-    | Cobol_config.Auto -> Cobol_preproc.Src_format.from_config SFFixed
-    | SF source_format -> Cobol_preproc.Src_format.from_config source_format
-  in
+  (* Note: this value doesn't actually matter, it will be overriden
+      immediately by [fold_source_lines] calling [on_initial_source_format]
+      below. *)
+  let src_format = Cobol_preproc.Src_format.from_config SFFixed in
   let state =
     Cobol_preproc.fold_source_lines ~dialect ~source_format
       ~on_initial_source_format:(fun src_format st -> { st with src_format })
--- a/src/lsp/cobol_indent/indenter.ml
+++ b/src/lsp/cobol_indent/indenter.ml
@@ -62,7 +62,7 @@ let indenter ~source_format (str:string) (rdl:indent_record list) range =
   String.concat "\n" strl
 
 (*indent a range of file, with the default indent_config*)
-let indent_range' ~dialect ~source_format ~range ~filename ~contents =
+let indent_range ~dialect ~source_format ~range ~filename ~contents =
   let state =
     Cobol_preproc.fold_source_lines ~dialect ~source_format
       ~skip_compiler_directives_text:true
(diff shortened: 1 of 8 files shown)


src/lsp/cobol_indent/indenter.ml
```diff
@@ -62,22 +72,37 @@ let indenter ~source_format (str:string) (rdl:indent_record list) range =
   String.concat "\n" strl
 
 (*indent a range of file, with the default indent_config*)
-let indent_range' ~dialect ~source_format ~range ~filename ~contents =
+let indent_range ~dialect ~source_format ~range ~filename ~contents =
+  let src_format =
```
nberth, line 76: Here you could just always start with `SFFixed` directly and not bother inspecting `source_format` at all.

src/lsp/cobol_indent/indenter.ml
```diff
@@ -62,22 +72,37 @@ let indenter ~source_format (str:string) (rdl:indent_record list) range =
...
       ~f:(fun _lnum line acc -> Indent_check.check_indentation line acc)
       (String { filename; contents })
-      { scope = BEGIN; context  = []; acc = []; range }
+      { src_format; scope = BEGIN; context  = []; acc = []; range }
   in
   (* NB: note here we ignore diagnostics *)
   let ind_recds = state.result.acc in
-  indenter ~source_format contents ind_recds state.result.range
+  indenter
+    ~source_format:state.result.src_format contents ind_recds state.result.range
```
nberth, line 100: Note, here it is the source format at the end of the source text that you are providing to `indenter`. Not sure this is intended (and what `indenter` uses it for).

nberth (commented):
I'm sure you meant "Only enable indenter on free format".

nberth:
You probably want to pass a `~toml:"source-format = \"fixed\"\n"` (the newline char may be mandatory) to calls to `make_lsp_project` in `test/lsp/lsp_formatting`.

src/lsp/cobol_indent/indenter.ml
bclement-ocp: True. This was initially written before #70, and I didn't revisit the code.

src/lsp/cobol_indent/indenter.ml
bclement-ocp: `indenter` uses this to know whether to indent or not :)

The indenter works in two passes — first it computes the changes using `fold_source_lines`, then it applies the changes separately. The first pass doesn't care about the format; the second pass uses the format to determine how to indent (e.g. ignore the first chars of the line).

I think the design is not ideal, because we end up with a single huge modification to the text where VSCode guidelines [state](https://code.visualstudio.com/api/language-extensions/programmatic-language-features):

> You should always return the smallest possible text edits that result in the source code being formatted. This is crucial to ensure that markers such as diagnostic results are adjusted correctly and are not lost.

It also assumes that there is a single source format per file, and I followed that assumption here -- this is why I use the source format at the end of the source text, so that the compiler directives are taken into account. If that assumption is wrong, which your reply makes me think it might be, some additional refactoring will be needed to deal with that.

bclement-ocp:
> You probably want to pass a `~toml:"source-format = \"fixed\"\n"` (the newline char may be mandatory) to calls to `make_lsp_project` in `test/lsp/lsp_formatting.ml`.

I'm sure you meant `~toml:"source-format = \"free\"\n"` ;) Done.

test/lsp/lsp_formatting.ml
```diff
@@ -23,7 +23,9 @@ let doc = {cobol|
   |cobol};;
 
 let%expect_test "simple-formatting-request" =
-  let { projdir; end_with_postproc }, server = make_lsp_project () in
+  let { projdir; end_with_postproc }, server =
+    make_lsp_project ~toml:"source-format = \"free\"\n" ()
```
nberth, line 27: This could have deserved a little `make_lsp_free_format_project` shortcut to avoid duplication.

nberth (approved):
I think you can address my comment if you wish, and then merge ;-)

test/lsp/lsp_formatting.ml
bclement-ocp: It could, but I believe test code deserves to be as plain as possible, hence I did not do that -- in fact there is much code duplication between the test cases already.

PR #75, opened 2023-10-26
Default superbol.path option

Stevendeo:
This PR sets the default value `superbol.path` to the executable name. This works when the `PATH` environment variable already points to the executable. This probably is not cross-compatible though.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
           "description": "If something is selected, only format the selection"
         },
         "superbol.path": {
-          "default": "",
+          "default": "superbol-free",
           "description": "Path to the `superbol` command"
         }
       }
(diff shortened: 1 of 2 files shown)


nberth:
The executable name should be `superbol-free` instead of `superbol` only I think.

Stevendeo:
An alternative would be to split the option in two:
- the PATH to the superbol binary;
- the name of the binary.

nberth:
> An alternative would be to split the option in two:
> 
> * the PATH to the superbol binary;
> * the name of the binary.

A slightly longer plan might be to ship the binary within the `vsix` extension file. So I'm not sure it is worth tweaking these configurations much further.

GitMensch:
This PR is good as is. I suggest to pull it in.

> A slightly longer plan might be to ship the binary within the `vsix` extension file. So I'm not sure it is worth tweaking these configurations much further.

I'm creating an issue for that.

PR #83, opened 2023-10-26
package.json outline for pattern matchers

GitMensch:
This is only an outline as the file is generated. Maybe @Stevendeo can take that into the original file with the correct indentation?

Not sure if we want to suffix all matchers with "-cobc" (first commit) or drop that (second).

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,6 +26,82 @@
         }
       }
     },
+        "problemPatterns": [
+            {
+                "name": "gnucobol-cobc",
+                "regexp": "^(.*): ?(\\d+): (error|warning): ([^[]*)(\\[(.*)\\])?$",
+                "file": 1,
+                "line": 2,
+                "severity": 3,
+                "message": 4,
+                "code": 6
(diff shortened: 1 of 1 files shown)


GitMensch:
... and not sure where the task definitions:

```json

    {
      "type": "superbol",
      "copybooks": [],
      "sourceFormat": "",
      "dialect": "",
      "forDebugging": false,
      "problemMatcher": [],
      "label": "superbol: Build file",
      
    },
    {
      "type": "superbol",
      "copybooks": [],
      "sourceFormat": "",
      "dialect": "",
      "forDebugging": true,
      "problemMatcher": [],
      "label": "superbol: Build file for debug"
    }
```

 are generated from but instead of `"problemMatcher": [],` it is then reasonable to generate `"problemMatcher": ["$gnucobol-error","$gnucobol-warning","$gnucobol-note","$gnucobol"],` .

nberth:
Changes included in #91 .

PR #87, opened 2023-10-26
initial gitpod configuration

GitMensch:
While this likely isn't "final" this seems to be the minimal needed setup.
I've started locally but recognized fast, that this won't work as the whole vsix part needs node 14 because of its dependencies (stable Debian distributes node 12, which I even had not installed) and the LSP part seems to need a complete ocaml environment.

Much nicer to have that online available.

I'll likely update that later and possibly add a link to use that to a hacking file (currently working on the README).

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,39 @@
+
+tasks:
+
+- name: setup GnuCOBOL and COBOLworx debugging environment on Ubuntu 22.04
+  init: |
+    mkdir -p _build/install/deb
+    pushd _build/install/deb
+    wget "https://cobolworx.com/pages/downloads/gnucobol_3.2.-1_amd64.deb"
+    wget "https://cobolworx.com/pages/downloads/cblgdb_4.28.4-1_amd64.deb"
+    pushd
+  before: |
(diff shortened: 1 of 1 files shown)


GitMensch:
@lefessan What's your take on this PR?

.gitpod.yml
```diff
@@ -0,0 +1,39 @@
...
+    # fixing broken interpreter in cobcd 
+    sudo sed -i -e 's#python2#python3#' /usr/bin/cobcd
+
+- name: setup coding environment on Ubuntu 22.04 (superbol-free)
+  before: |
+    sudo apt install opam ocaml-dune
+    opam init
+    # note autofonce-lib currently needs to be manually installed, see https://github.com/OCamlPro/superbol-studio-oss/issues/79
+    opam install autofonce_lib
+    opam install ./opam/*.opam -y--deps-only
```
nberth, line 25: I think here that would now be `make dev-deps` in place of the two calls to `opam install`.

nberth (commented):
I've probably missed many aspects, and don't even know how to test this (yet). But that could be an interesting addition indeed.

Issue #76, opened 2023-10-26
FR: distribute the language-server along with the .vsix


GitMensch:
> A slightly longer plan might be to ship the binary within the `vsix` extension file.

_Originally posted by @nberth in https://github.com/OCamlPro/superbol-studio-oss/issues/75#issuecomment-1781089922_

This issue is about doing so, using platform-specific extensions.

More details are found at https://code.visualstudio.com/api/working-with-extensions/publishing-extension#platformspecific-extensions and of course - for the client side - https://github.com/microsoft/vscode-vsce.

There would be one version for each "target" you want plus one extra without a target (which would need an external install and possibly configuration to reach it as it is now).

bclement-ocp:
Do we need platform-specific extensions if we ship a version of the language server compiled to JavaScript?

GitMensch:
No. That won't be necessary. But I don't know if/how that affects the performance.

Compiled to javascript may also be useful for vscode for web (without a VM).

nberth:
Closing this as it has been completed by #124 .

Web-specific version will be discussed in a dedicated issue.

PR #93, opened 2023-10-27
Provide a "universal" VSIX, with the LSP server compiled in JavaScript

bclement-ocp:
This is a first step towards fixing #76 (but does not entirely fixes it).  It avoids the need to install an external superbol executable, and should work on all platforms.

The bundled superbol-free.bc.js is used if the option `superbol.path` is set to `null`, which is the new default. Some performance investigation is needed -- if we find out that the js_of_ocaml version is too slow for practical use, we can make this not be the default.


src/lsp/superbol-free/dune
nberth, line 1: Please investigate whether that can be achieved via fields of the corresponding `package.toml`. Same for `superbol_free_lib/dune`.

Makefile
nberth, line 1: This file is generated by `drom` from `Makefile.drom-tpl`. Please edit that other file instead.

package.json
```diff
@@ -21,8 +21,12 @@
...
         },
         "superbol.path": {
-          "default": "superbol-free",
-          "description": "Path to the `superbol` command"
+          "type": [
+            "string",
+            "null"
+          ],
+          "default": null,
+          "description": "Path to the external `superbol` command. If `null`, use the bundled `superbol-free`."
```
nberth, line 29: Note this conflicts with pending edits in #72 .

nberth (changes requested):
This seems quite promising. We'll just have to check the impacts on performance (we can see to edit some of the NIST85 sources to check that :wink: — to be configured with `dialect = "cobol85"`).

We'll refrain from merging right now, though: I'd rather let `drom` continue to automatically manage/generate some of the files edited here.

src/lsp/superbol-free/dune
bclement-ocp: I have changed the drom files. Running `drom build` generates a bunch of `main.ml` files everywhere tho, not sure what's up about that. Not included in this commit, in spite of drom "helpfully" (not at all) deciding to add its changes to the index.

nberth: Might depend on `drom`'s version. I typically use `opam exec -- drom project` to generated the files instead.

src/vscode/superbol-vscode-platform/dune
nberth, line 1: I think in this case there is a `dune.drom-tpl` to edit instead.

bclement-ocp: I don't think this is a change I made — `drom` modified this file.

Makefile.header
```diff
@@ -15,35 +24,59 @@ superbol-free: build
...
+                --sourcemap
+# the last command generated _dist/$(CLIENT_PROJECT).bs.js
 
 .PHONY: build-release yarn-release
 
-build-release:
-	opam exec -- dune build $(SRCDIR)/$(PROJECT).bc.js --profile=release
-	mkdir -p _out
-	$(CP) _build/default/$(SRCDIR)/$(PROJECT).bc.js _out/
+build-release: _out/$(CLIENT_PROJECT).bc.js
```
nberth, line 64: Why not `_out/$(SERVER_PROJECT).bc.js` as well?

bclement-ocp: Good catch, I simply forgot :)

src/vscode/superbol-vscode-platform/dune
nberth: I saw something related to `js_of_ocaml` so I though it was edited manually. We should probably check those lines were not introduced on purpose.

bclement-ocp:
FWIW, I have checked on a bunch of files from NIST85 and the performance difference, if any, is not noticeable on my machine — except that with the `js_of_ocaml` version I get stack overflows when computing semantic tokens on large files, which is an issue.

This is probably due to tail recursion not being handled properly by `js_of_ocaml` outside of mutually recursive nests, and I guess we have tail recursion that doesn't fall in this category (some CPS stuff maybe?). I have tried the `--enable=effects` flag from `js_of_ocaml` but there the performance difference is sadly noticeable :(

nberth:
> FWIW, I have checked on a bunch of files from NIST85 and the performance difference, if any, is not noticeable on my machine — except that with the `js_of_ocaml` version I get stack overflows when computing semantic tokens on large files, which is an issue.

Yes I just made the same observations… do you think adding some `tailcall` annotations may help identify where the issues could be?

bclement-ocp:
> > FWIW, I have checked on a bunch of files from NIST85 and the performance difference, if any, is not noticeable on my machine — except that with the `js_of_ocaml` version I get stack overflows when computing semantic tokens on large files, which is an issue.
> 
> Yes I just made the same observations… do you think adding some `tailcall` annotations may help identify where the issues could be?

As far as I know the `@tailcall` annotation requires the compiler to fail if the the call can't be optimized, it won't help here — js_of_ocaml doesn't perform tail call elimination (except in very specific cases, cf https://ocsigen.org/js_of_ocaml/latest/manual/tailcall )

nberth:
> As far as I know the `@tailcall` annotation requires the compiler to fail if the the call can't be optimized, it won't help here — js_of_ocaml doesn't perform tail call elimination (except in very specific cases, cf https://ocsigen.org/js_of_ocaml/latest/manual/tailcall )

Ah ok. That was in case that annotation could also be checked by `js_of_ocaml` itself. It seams only some LSP requests appear to suffer from the issue… is it possible to add `--enable=effects` to only some libraries? or is it a flag that only applies globally?

bclement-ocp:
`js_of_ocaml` takes bytecode as input — I don't think that the annotation is preserved until there in the compilation chain. `--enable=effects` applies a CPS transformation and I believe already tries to keep code in direct style when possible, so I think we are kind of stuck with the #76 suggestion of using platform-dependent extensions for now. Will discuss with @ddeclerck what is the best way to build those.

src/vscode/superbol-vscode-platform/dune
bclement-ocp: We should ask @ddeclerck they were introduced in a commit "fix cross-compilation" (09b9ce7c5d791a5d348337cc7d152780b13e7a56)

bclement-ocp:
Marking as draft because it seems I made a mistake somewhere — I thought this was using the bundled `node` from vscode but it actually is not, so I will have to investigate more, if we do want to do this.

src/vscode/superbol-vscode-platform/dune
ddeclerck: This `dune.drom-tpl` is based on `https://github.com/OCamlPro/drom-share/blob/master/packages/js_program/dune_`. I had to add it so I could add a missing `!(package-dune-stanzas)` (which comes from the `package.toml` file and disables compiling the VSCode extension when cross-compiling : js_of_ocaml is neither available nor needed in cross-compilation environments). Obviously, this is a temporary solution ; the template dune file should be modified in the aforementioned `drom-share` repository. If someone feels like contributing to `drom`.

GitMensch:
Several changes done here (at least the parts that change hard-wired entries to variables) seem reasonable in any case. Can those get it with a direct commit/separate PR so that the actual "javascript server" PR will include less changes?

nberth:
Labeling this as history. Performance issues with the Javascript+effects version seem to be too hard to deal with at the moment. Another way towards a "pure" web extension could be to compile the LSP server into WASM using `wasm_of_ocaml` (or another such compiler).

nberth:
Reopening after [more investigations](#384)…

.gitattributes
```diff
@@ -5,3 +5,4 @@ linking_flags.sh linguist-generated
 /dune-project    linguist-generated
 /sphinx/conf.py  linguist-generated
 /package.json    linguist-generated
+/**/version.mlt  linguist-generated
```
nberth, line 8: @emilienlemaire Could you move these changes to a separate PR? (This kind of mismatch should not happen — caused by some hasty change in `drom`).

emilienlemaire: Just this one or all the changes from all `version.mlt` as well?

Makefile.drom-tpl
```diff
@@ -32,8 +32,12 @@ build:
 	./scripts/before.sh build
 ifeq ($(TARGET_PLAT)_$(BUILD_STATIC_EXECS),linux_true)
 	./scripts/static-build.sh
+else
+ifeq ($(TARGET_PLAT), web)
```
nberth, line 36: Take care of the space before `web`; I think it's better to remove it.

Makefile.header
```diff
@@ -19,17 +19,27 @@ DUNE_BUILD_SRCDIR =							\
...
+  TARGET_SPEC = $(TARGET_PLAT)
+else
+  TARGET_SPEC = $(TARGET_PLAT)-$(TARGET_ARCH)
+endif
+
 TARGET_PROFILE ?= release
 TARGET_VSIX = $(TARGET_NAME)-$(VERSION)-$(TARGET_SPEC)-$(TARGET_PROFILE).vsix
 
 DOT_EXE = $(if $(filter win32,$(TARGET_PLAT)),.exe)
 SUPERBOL_LSP = superbol-free-$(TARGET_SPEC)$(DOT_EXE)
```
nberth, line 32: Can you move this in the `else` case below (for clarity).

opam/cobol_data.opam
```diff
@@ -47,6 +47,7 @@ install: [
 depends: [
   "ocaml" {>= "4.14.0"}
   "dune" {>= "2.8.0"}
+  "zarith_stubs_js" {>= "0.17"}
```
nberth, line 50: Hum… we need to investigate what impact that has. Ideally we'd find a way to avoid this dependency in the `opam` file (or have a flag that says the dependency is only valid for `js` target — not sure this is doable)?

src/vscode/superbol-vscode-platform/superbol_languageclient.ml
```diff
@@ -48,7 +48,8 @@ let find_superbol root =
     Vscode.Uri.joinPath root ~pathSegments:[name]) @@ [
     Format.asprintf "%s-%s-%s%s" prefix platform arch suffix;
     Format.asprintf "%s-%s%s" prefix platform suffix;
-    Format.asprintf "%s%s" prefix suffix
+    Format.asprintf "%s%s" prefix suffix;
+    Format.asprintf "main.bc.js"
```
nberth, line 52: Here that comes before the `arm64`->`x64` special case. I guess we'd need that further down the list.

.gitattributes
nberth: Ideally this one and the changes to `version.mlt`s.

opam/cobol_data.opam
emilienlemaire: None I think, it only contains a `.js` file, so it won't be used unless compiled in JS mode

src/vscode/superbol-vscode-platform/superbol_languageclient.ml
emilienlemaire: Good catch, I'll add it after the last check

nberth (approved):

PR #94, opened 2023-10-27
Support change of reference format while indenting

bclement-ocp:
We now use the current source format to indent each line.

Related cleanup: the indenter now uses an array rather than a list to represent the content of the file for performance.

Note: currently includes #93.


nberth (approved):
Ok to merge once the CI terminates.

PR #95, opened 2023-10-27
Instruct github about some more auto-generated files

nberth:
Github shows changes to many files that are actually generated by `drom`. The new `.gitattributes` instructs it to hide the diffs by default.

Note: some of the generated files might still be relevant for review, and some others might be missing. So suggestions are welcome.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,7 @@
+dune             linguist-generated
+linking_flags.sh linguist-generated
+/.drom           linguist-generated
+/Makefile        linguist-generated
+/dune-project    linguist-generated
+/sphinx/conf.py  linguist-generated
+/package.json    linguist-generated


bclement-ocp (approved):
This looks sensible. Maybe we could also consider having a CI workflow to check that those generated files are up-to-date (and possibly automatically re-generate them in a new commit)?

ddeclerck:
Maybe you could have a single file with the list of generated files and their hashes, so as the reviewer at least "knows" they have changed, and check the diff only if he needs to. Now that I wrote this, I realize this is just the `.drom` file ;)

bclement-ocp:
> Maybe you could have a single file with the list of generated files and their hashes, so that the reviewer at least "knows" they have changed, and check the diff only if he needs to. Now that I wrote this, I realize this is just the `.drom` file ;)

The way `linguist-generated` files are handled by the GitHub UI is that they are folded by default but still included in the "Review" tab (we were confused by this over on the Alt-Ergo repository), in a similar way as for large files, so this information would be in the UI already :)

ddeclerck:
> > Maybe you could have a single file with the list of generated files and their hashes, so that the reviewer at least "knows" they have changed, and check the diff only if he needs to. Now that I wrote this, I realize this is just the `.drom` file ;)
> 
> The way `linguist-generated` files are handled by the GitHub UI is that they are folded by default but still included in the "Review" tab (we were confused by this over on the Alt-Ergo repository), in a similar way as for large files, so this information would be in the UI already :)

Then there is in fact no problem :p

PR #96, opened 2023-10-27
Prevent drom from generating main.ml

bclement-ocp:
Not sure why it is generating main.ml for library projects, but I don't want them, so I add them to the `skip` field in `drom.toml` as instructed by the generated files themselves.

(Using drom 0.9.0)

### DIFF
--- a/drom.toml
+++ b/drom.toml
@@ -33,7 +33,7 @@ of the superbol-studio-oss OCaml project
 """
 
 [project]
-skip = ["@test", "@ocamlformat", "@ocp-indent", "README.md", "sphinx/about.rst", "LICENSE.md", "sphinx/index.rst"]
+skip = ["@test", "@ocamlformat", "@ocp-indent", "README.md", "sphinx/about.rst", "LICENSE.md", "sphinx/index.rst", "src/lsp/pretty/main.ml", "src/lsp/cobol_typeck/main.ml", "src/lsp/cobol_ptree/main.ml", "src/lsp/cobol_lsp/main.ml", "src/lsp/cobol_lsp/main.ml", "src/lsp/cobol_data/main.ml", "src/lsp/cobol_indent/main.ml", "src/testing/superbol_testutils/main.ml"]
 
 # project-wide library dependencies (not for package-specific deps)
 [dependencies]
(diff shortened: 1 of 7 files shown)


drom.toml
```diff
@@ -33,7 +33,7 @@ of the superbol-studio-oss OCaml project
 """
 
 [project]
-skip = ["@test", "@ocamlformat", "@ocp-indent", "README.md", "sphinx/about.rst", "LICENSE.md", "sphinx/index.rst"]
+skip = ["@test", "@ocamlformat", "@ocp-indent", "README.md", "sphinx/about.rst", "LICENSE.md", "sphinx/index.rst", "src/lsp/pretty/main.ml", "src/lsp/cobol_typeck/main.ml", "src/lsp/cobol_ptree/main.ml", "src/lsp/cobol_lsp/main.ml", "src/lsp/cobol_lsp/main.ml", "src/lsp/cobol_data/main.ml", "src/lsp/cobol_indent/main.ml", "src/testing/superbol_testutils/main.ml"]
```
nberth, line 36: I myself had many more entries here not so long ago… wondering what happened.

I'll try to find them.

nberth: Actually that was in an earlier version/different branch of this project.

bclement-ocp: Maybe we can import the `skip` list from that earlier version then?

nberth: Could you run `opam exec -- drom project` to see if it still generates some files we don't want? If no, then that PR  can probably be merged as is.

bclement-ocp:
I have updated this MR with the current result of running `drom project`. Looks like `drom` is modifying some other files as well as ignoring some modifications to `Makefile.drom-tpl`. I'll let you merge it or not as appropriate…

nberth:
When I see the changes you get from `drom`, my hunch is you are not using *the* proper version of `drom`. At the current time you need to pin `drom`'s repository to get it running properly `opam pin add https://github.com/OCamlPro/drom.git` on SuperBOL.

bclement-ocp:
You are right: I was using version 0.9.0, as I understood was the version to use from earlier meetings. Turns out that it's not. `drom` no longer generates those files when using commit `63a5770b0a94826675b588677ba38316409eac15`.

That might deserve a mention in a README somewhere.

nberth:
> That might deserve a mention in a README somewhere.

Yes I'm currently updating the docs, notably to help newcomers.

Issue #46, opened 2023-10-10
Move auto-indent configuration to per-project file


lefessan:
This must done after having moved the LSP per-project configuration to a generic per-project file (#47)

bclement-ocp:
Taking this one.

PR #97, opened 2023-10-30
Use minimal diffs for the indenter

bclement-ocp:
Internally, the indenter computes a diff (in terms of offsets) for each line that needs to change. Then, all the diffs are applied to the document, and a single edit that rewrites the full document is sent to the editor.

This PR changes the behavior to send small text edits that either add or remove whitespace instead.

### DIFF
--- a/src/lsp/cobol_indent/cobol_indent.ml
+++ b/src/lsp/cobol_indent/cobol_indent.ml
@@ -16,61 +16,9 @@ module Type = Indent_type
 (*return the result of indentation. use user-defined indent_config*)
 let indent_range = Indenter.indent_range
 
-(** [apply s rdl] applies the changes in [rdl] to the text in [s]. [rdl] is
-    expected to have been computed by calling [indent_range ~contents:s]. *)
-let apply : string -> Type.indent_record list -> string =
-  (*
-    [irs] contains a list of {!Type.indent_record} entries describing how to
-     modify the indentation.  They must be sorted in increasing order, and there
-     must be at most one record per line.
-
-    [buf] is a {!Buffer.t} into which we are writing the indented content.
-
-    [s] is the original content to indent as a {!string}.
-
-    [line] is the current line, which starts at offset [bol] in the string [s].
-
-    [start] is an offset in [s], less than or equal to [bol], that must be
-    preserved as-is. This allows copying large unmodified chunks of text (lines
-    that are already properly indented) using a single call to
-    {!Buffer.add_substring} for reduced overhead.
-  *)
-  let rec aux ~start (irs : Type.indent_record list) buf s line bol =
-    (* Note: this can raise [Invalid_argument] if the modifications do not
-       cleanly apply to the input string, which include cases where there are
-       modifications on lines past the end of the input, modifications that are
-       not sorted, etc. -- we can call [invalid_arg] ourselves, but
-       [String.index_from] can also raise [Invalid_argument] in these
-       conditions. *)
-    match irs with
-    | [] when start = 0 -> s
-    | [] ->
-      Buffer.add_substring buf s start (String.length s - start);
-      Buffer.contents buf
-    | { lnum; offset_orig; offset_modif } :: irs when line = lnum - 1 ->
-      let eol =
-        try String.index_from s bol '\n' + 1 with Not_found -> String.length s
-      in
-      let delta = offset_modif - offset_orig in
-      if delta > 0 then (
-        Buffer.add_substring buf s start (bol + offset_orig - start);
-        for _ = 0 to delta - 1 do
-          Buffer.add_char buf ' '
-        done;
-      ) else (
-        Buffer.add_substring buf s start (bol - start + offset_orig + delta);
-      );
-      aux ~start:(bol + offset_orig) irs buf s (line + 1) eol
-    | _ :: _ ->
-      match String.index_from s bol '\n' + 1 with
-      | eol -> aux ~start irs buf s (line + 1) eol
-      | exception Not_found ->
-        (* Modifications on lines past the end of the input. *)
-        invalid_arg "Cobol_indent.apply"
-  in
-  fun s rdl ->
-    try
-      let buf = Buffer.create (String.length s) in
-      aux ~start:0 (List.rev rdl) buf s 0 0
-    with Invalid_argument _ ->
-      Fmt.failwith "Could not indent: internal error"
+let indent_range_str
+  ~dialect ~source_format ~indent_config ~range ~filename ~contents
+=
+indent_range
+  ~dialect ~source_format ~indent_config ~range ~filename ~contents
+|> Indent_util.apply contents
--- a/src/lsp/cobol_indent/cobol_indent.ml
+++ b/src/lsp/cobol_indent/cobol_indent.ml
@@ -18,7 +18,7 @@ let indent_range = Indenter.indent_range
 
 (** [apply s rdl] applies the changes in [rdl] to the text in [s]. [rdl] is
     expected to have been computed by calling [indent_range ~contents:s]. *)
-let apply =
+let apply : string -> Type.indent_record list -> string =
   (*
     [irs] contains a list of {!Type.indent_record} entries describing how to
      modify the indentation.  They must be sorted in increasing order, and there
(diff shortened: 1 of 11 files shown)


src/lsp/cobol_indent/indent_config.ml
```diff
@@ -62,9 +62,11 @@ let () =
       "SECTION", 0 ]
 
 let set_config ~indent_config =
-  let str = Ez_file.V1.EzFile.read_file indent_config in
-  let strlist = String.split_on_char '\n' str in
-  build_table strlist offset_table
+  match Ez_file.V1.EzFile.read_file indent_config with
```
nberth, line 65: Please note: this should be temporary, and the config for the indenter should come as a pre-filled record for the `cobol_indent` library (typically, `indent_config.ml`, bound as `Cobol_indent.Config`, would only define the type of this record).

src/lsp/cobol_indent/cobol_indent.ml
```diff
@@ -16,10 +16,62 @@ module Type = Indent_type
...
-let indent_file ~dialect ~source_format ~file ~indent_config =
-  let contents = Ez_file.V1.EzFile.read_file file in
-  indent_range
-    ~dialect ~source_format ~range:None ~indent_config
-    ~filename:file ~contents
-  |> Fmt.pr "%s"
\ No newline at end of file
+(** [apply s rdl] applies the changes in [rdl] to the text in [s]. [rdl] is
+    expected to have been computed by calling [indent_range ~contents:s]. *)
+let apply =
```
nberth, line 21: If you don't name the arguments here, could you add the type annotation? Then they'd stand out better than in the comment.

src/lsp/cobol_indent/cobol_indent.ml
```diff
@@ -16,10 +16,62 @@ module Type = Indent_type
...
+  let rec aux ~start (irs : Type.indent_record list) buf s line bol =
+    (* Note: this can raise [Invalid_argument] if the modifications do not
+       cleanly apply to the input string, which include cases where there are
+       modifications on lines past the end of the input, modifications that are
+       not sorted, etc. -- we can call [invalid_arg] ourselves, but
+       [String.index_from] can also raise [Invalid_argument] in these
+       conditions. *)
+    match irs with
+    | [] ->
+      if start = 0 then s else (
```
nberth, line 47: I'd use a `[] when start = 0 -> s | [] -> ...` for better readability.

src/lsp/cobol_indent/cobol_indent.ml
```diff
@@ -16,10 +16,62 @@ module Type = Indent_type
...
+        Buffer.add_substring buf s start (String.length s - start);
+        Buffer.contents buf
+      )
+    | { lnum; offset_orig; offset_modif } :: irs when line = lnum - 1 ->
+      let eol =
+        try String.index_from s bol '\n' + 1 with Not_found -> String.length s
+      in
+      let delta = offset_modif - offset_orig in
+      if delta > 0 || true then (
```
nberth, line 56: Why `|| true` (which kills `else` branch)?

nberth (commented):
Overall fine. Just one or two questions mostly.

src/lsp/cobol_indent/cobol_indent.ml
bclement-ocp: Because this is a leftover from some debugging, good catch (note that this code is only used in the `superbol indent` command).

src/lsp/cobol_indent/cobol_indent.ml
bclement-ocp: Thought I did that actually, will change.

src/lsp/cobol_indent/cobol_indent.ml
bclement-ocp: Sure!

src/lsp/cobol_indent/indent_config.ml
bclement-ocp: Ah, this is #46 — I understand that issue now. Yeah, this is just a hotfix to be able to test stuff in spite of the hard-coded path used in `superbol indent`.

src/lsp/cobol_lsp/lsp_request.ml
```diff
@@ -168,16 +181,7 @@ let handle_range_formatting registry params =
...
-  let range =
-    (*TODO:find a simple method to get the number of letters of one line
-           (the code below use 1000 as the upperbound)                   *)
-    Range.{
-      start = Position.create ~line:start.line ~character:0;
-      end_ = Position.create ~line:end_.line ~character:1000;
-    }
-  in
-  let newText =
+  let editList =
```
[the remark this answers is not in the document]
bclement-ocp: I just kept the local style (`newText` was named this way because it is passed to `TextEdit.create` which expects a `newText` named parameter, I believe). No issues with using snake case (but this is O**Caml**!).

src/lsp/cobol_indent/cobol_indent.ml
```diff
@@ -43,17 +43,16 @@ let apply =
...
-      )
+      Buffer.add_substring buf s start (String.length s - start);
+      Buffer.contents buf
     | { lnum; offset_orig; offset_modif } :: irs when line = lnum - 1 ->
       let eol =
         try String.index_from s bol '\n' + 1 with Not_found -> String.length s
       in
       let delta = offset_modif - offset_orig in
-      if delta > 0 || true then (
+      if delta > 0 then (
```
nberth, line 55: Are you sure this is the intended change? I'd got with simply removing the condition and only keeping the `then` branch, rather than trying to optimize the loop out in case `delta ≤ 0` (with a comment "no-op if `delta <= 0`): I personally think that, in general, less lines of code is usually preferable.

bclement-ocp: When `delta < 0` it's not a no-op, we do need to remove the spaces.

bclement-ocp: I mean we do need to remove the characters, we don't know that they are spaces at this point (but the indenter will only remove actual spaces).

nberth: Ok I see now. The `|| true` made me think the other case was not needed as all the tests seemed to pass). What this actually reveals is that we miss at least one test where the indenter actually need to un-indent.

Please could you add at least one that exercises the second branch? In `test/lsp/lsp_formatting.ml` might be a good place.

bclement-ocp: As I mentioned above we actually don't test this at all. It is only used by the `superbol indent` command, *not* by the LSP (in the LSP the actual edits are made by the editor using `TextEdit`, and in the tests we test the LSP interface with the `Lsp.Text_document.apply_text_document_edits` method — but `Cobol_indent` does not depend on the LSP.

That said — we indeed have no tests where the indenter has to de-indent. Added one.

nberth: Yes we don't actually test `superbol indent` at all (or the relevant portions of `Cobol_indent`). Then, to finalize this PR, I suggest to proceed as follows:
- move the `apply` function out of `cobol_indent.ml` (where we should typically have a minimal set of module bindings, and if well justified for the library API, module inclusions — the current contents of this file does not follow this rule yet), maybe in a `indent_string.ml` — no need for a module binding in `cobol_indent.ml`, unless it is accessed from outside of the library;
- add a comment around `apply` that it is not tested yet.

Then I think this PR can be merged.

bclement-ocp: Well it is used by `command_indent_file.ml` in `superbol_free_lib`. Although I can easily introduce a wrapper function `indent_range_to_string` or some such instead.

(My personal opinion is that we shouldn't have this code path at all and should just have `cobol_indent` depend on `lsp` but I get that we may not want that)

nberth: Yes, that'd be quite a large dependency for such a library… so it's debatable.

If the value is used then I'd either bind the new module directly so one can `Cobol_indent.String.apply`, or else just the value like `Cobol_indent.apply_on_string`, or something similar. The idea is to end-up with a clean interface for the library, with function names that are not too redundant as they should typically be called with full qualification.

bclement-ocp: I exposed the `indent_range_str` instead. `apply` is only indented to be used on the result of `indent_range` anyways.

> Yes, that'd be quite a large dependency for such a library… so it's debatable.

I think the point is moot, because realistically there will be no project that depend on `Cobol_indent` and not the rest of Superbol. Anyways, the code exists and works, so it doesn't really matter.

nberth: > there will be no project that depend on `Cobol_indent` and not the rest of Superbol.

I wouldn't be so sure. One may want a "small" executable that only focuses on indentation (like `ocp-indent`), that some other editor like GNU/Emacs can use in place of the full-blown LSP for "lighter" tasks. For instance, OCaml I use a combination of `merlin` and `ocp-indent`, because I find that combination more customizable.

Anyways, `cobol_indent` is a small library, so I think we can merge at this point.

nberth (approved):
Fine ;-)

src/lsp/cobol_indent/cobol_indent.ml
bclement-ocp: This feels academic since the design of the project is to provide a single super-charged executable with subcommands though; most people would probably just use `superbol indent`.

nberth: Now that I think about it, the code for effectively indenting a file should just not belong to `cobol_indent`; there, we should just stick to the indentation logic/algorithm. By that approach, `Superbol_free_lib` may just rely on `lsp` to apply the text editions.

(Apologies if there was an `indent_file` function already there; it should have been removed before).

PR #98, opened 2023-10-30
Bugfix: lookup of pre-processor reset position upon rewind

nberth:
The commit that matters most is 227b3689:

>    This fixes a bug introduced in 86e6cca6 by an uncareful optimization
>    of the computation.  Essentially, the meaning of `Failure` and
>    `Invalid_argument` exception in `List.nth` needed to be reversed.  The
>    included changes fix and improve the affected portion of code, and add
>    some new (high-level) tests to better exercise the rewinding code.

### DIFF
--- a/src/lsp/cobol_preproc/preproc_engine.ml
+++ b/src/lsp/cobol_preproc/preproc_engine.ml
@@ -55,25 +55,27 @@ let rev_ignored { reader; _ } = Src_reader.rev_ignored reader
 
 (** [position_at ~line ~char pp] computes a lexing position that corresponds to
     the given line and character indexes (all starting at 0) in the input
-    already read by [pp].  Raises {!Not_found} if no complete line was processed
-    yet, or the current position if the given line index doed not correspond to
-    already processed lines. *)
+    already read by [pp].  Raises [Not_found] if no complete line was processed
+    yet, or the current position if the given line index does not correspond to
+    an already processed line. *)
 let position_at ~line ~char { reader; _ } =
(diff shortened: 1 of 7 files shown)


src/lsp/cobol_preproc/preproc_engine.ml
```diff
@@ -55,25 +55,27 @@ let rev_ignored { reader; _ } = Src_reader.rev_ignored reader
...
+  else     (* |rev_newline_cnums| is the number of newline chars seen upto... *)
+    let lexpos = Src_reader.position reader in       (* ... current position. *)
+    if line <= lexpos.pos_lnum - 1                   (* = |rev_newline_cnums| *)
+    then
       let pos_bol =
-        try List.nth rev_newline_cnums (lexpos.pos_lnum - line - 1)
-        with Not_found | Invalid_argument _ -> 0
+        if line <= 0
+        then 0
+        else List.nth rev_newline_cnums (lexpos.pos_lnum - 1 - line)
```
bclement-ocp, line 72: If I understand correctly, this can no longer fail because:
 - We check `line <= lexpos.pos_lnum - 1` above, so `lexpos.pos_lnum - 1 - line` cannot be negative; and
 - The length of `rev_newline_cnums` is equal to `lexpos.pos_lnum` (type invariant)

bclement-ocp (approved):

src/lsp/cobol_preproc/preproc_engine.ml
nberth: Yes that's about it.

Issue #81, opened 2023-10-26
(re)start LSP after change of `superbol.path`


GitMensch:
Currently one needs to reload all related windows.

PR #99, opened 2023-10-30
Add "Superbol: Restart Language Server" command.

### LINKED ISSUE #88: Add Superbol: Restart Language Server command
We should be able to restart the language server when it gets stuck in a buggy state (similar to what is available on the OCaml Language Server).

Related: #81

bclement-ocp:
Add a command to restart the language server. This will pick up changes to the `superbol.path` configuration option.

Fixes #88

Note that this allows a manual solution to #81 (making it possible to start the restart language server after changing `superbol.path` without having to reload the whole extension).

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,13 @@
         "language": "COBOL"
       }
     ],
+    "commands": [
+      {
+        "command": "superbol.server.restart",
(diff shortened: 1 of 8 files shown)


nberth (approved):
Looks good to me!

PR #103, opened 2023-10-30
Use per-project file for indent configuration

### LINKED ISSUE #46: Move auto-indent configuration to per-project file

bclement-ocp:
The indent configuration is now exposed by the `Cobol_indent` module and loaded from the `superbol.toml` file. For the sake of simplicity, this does not change the internal representation of the indent configuration, even though it requires us to expose the configuration as a hash table in `Indent_type.ml`. Be responsible.

Fixes #46


nberth (commented):
Mostly fine ;-). As usual just one or two suggestions.

nberth:
Actually, I think we can merge as is. The suggestions are very minor and can be implemented in subsequent PRs.

PR #109, opened 2023-11-02
Fix per-file project-loading

nberth:
Includes fixes to project directory lookup for `indent` sub-commands, as well as some basic renaming of modules to have a meaningful `Superbol_free_lib.Project` module.


bclement-ocp (commented):
The files `vSCode_extension.ml{,i}` should probably be called `vscode_extension.ml{,i}` instead. `vSCode_extension.ml{,i}` is really ugly; `VSCode_extension.ml{,i}` would be fine, but there are other references to `Vscode` using this case (e.g. `Vscode_json`, `Vscode_languageclient`) and consistency matters.

nberth:
> The files `vSCode_extension.ml{,i}` should probably be called `vscode_extension.ml{,i}` instead. `vSCode_extension.ml{,i}` is really ugly; `VSCode_extension.ml{,i}` would be fine, but there are other references to `Vscode` using this case (e.g. `Vscode_json`, `Vscode_languageclient`) and consistency matters.

Yes, I am planning to rename all of those other ulgy file names. The name of the module matters (and should be VSCode`), not the name of the file.

nberth:
I can  change to `VSCode_extension.ml{,i}` if you  prefer.

bclement-ocp:
We will disagree here because I think that using `Vscode` rather than `VSCode` as the module name is also the most consistent with the rest of the project using `snake_case`, including for constructor names (not having to worry about whether a capital should be added in a module name or not is nice), but I am fine with any solution that doesn't involve `vSCode` (which is introduced in the PR, if I'm not mistaken).

nberth:
Hum, I always though VSCode was written like this (but I don't use it myself so I never checked).
Actually it seems "VS Code" (with a space) is the way it is spelled, which indicates `VS_code_extension` should be the module name (acronyms should stay in capitals even in module names). But as you point out we already have a name for that in the project so I'll keep `Vscode_`.

PR #114, opened 2023-11-06
Look for bundled superbol-free executable

bclement-ocp:
This patch makes the VS Code extension look for a bundled executable in the following order (and with a `.exe` suffix on Windows):

 - `_dist/superbol-free-${platform}-${arch}`
 - `_dist/superbol-free-${platform}`
 - `_dist/superbol-free`

So for instance on x64 Linux we will look for:

 - `_dist/superbol-free-linux-x64`
 - `_dist/superbol-free-linux`
 - `_dist/superbol-free`

On recent Macs we will look for:

 - `_dist/superbol-free-darwin-arm64`
 - `_dist/superbol-free-darwin`
 - `_dist/superbol-free`

And on Windows we will look for

 - `_dist/superbol-free-win32-x64.exe`
 - `_dist/superbol-free-win32.exe`
 - `_dist/superbol-free.exe`

This will also allow tackling #76 once we actually put the executables in the extension. It also allows some flexibility between making "slim" per-platform extensions or "fat" multiplatform extensions with many executables.

### DIFF
--- a/src/vscode/superbol-vscode-platform/superbol_instance.ml
+++ b/src/vscode/superbol-vscode-platform/superbol_instance.ml
@@ -20,6 +20,8 @@ type t = {
 }
 
 let make () = {
+  (* If no bundled superbol is found, try superbol-free from PATH as
+     last-resort. *)
   bundled_superbol = "superbol-free";
   language_client = None
 }
@@ -38,7 +40,8 @@ let start_language_server t =
   let open Promise.Syntax in
   let* () = stop_language_server t in
   let serverOptions =
-    Superbol_languageclient.serverOptions t.bundled_superbol
+    Superbol_languageclient.serverOptions
+      ~bundled_superbol:t.bundled_superbol
   in
   let clientOptions = Superbol_languageclient.clientOptions () in
   let client = LanguageClient.make
--- a/src/vscode/superbol-vscode-platform/superbol_instance.ml
+++ b/src/vscode/superbol-vscode-platform/superbol_instance.ml
@@ -19,10 +19,8 @@ type t = {
   mutable language_client: LanguageClient.t option
 }
 
-let make () = {
-  (* If no bundled superbol is found, try superbol-free from PATH as
-     last-resort. *)
-  bundled_superbol = "superbol-free";
+let make ~bundled_superbol () = {
+  bundled_superbol;
   language_client = None
 }
 
@@ -50,6 +48,3 @@ let start_language_server t =
     ~serverOptions ~clientOptions () in
   let+ () = LanguageClient.start client in
   t.language_client <- Some client
-
-let set_bundled_superbol t bundled_superbol =
-  t.bundled_superbol <- bundled_superbol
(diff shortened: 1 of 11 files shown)


src/vscode/superbol-vscode-platform/superbol_languageclient.ml
```diff
@@ -12,13 +12,18 @@
 (*                                                                        *)
 (**************************************************************************)
 
-let serverOptions () =
+let serverOptions superbol_path =
```
nberth, line 15: Please could you label this argument? Otherwise it's hard to guess what is expected by `Superbol_languageclient.serverOptions`…

src/vscode/superbol-vscode-platform/superbol_instance.ml
```diff
@@ -15,10 +15,12 @@
 open Vscode_languageclient
 
 type t = {
+  mutable bundled_superbol : string;
   mutable language_client: LanguageClient.t option
 }
 
 let make () = {
+  bundled_superbol = "superbol-free";
```
nberth, line 23: Please can you comment on the justifications for having this as default?

nberth (commented):
Looks fine to me. I just need two little clarifications.

src/vscode/superbol-vscode-platform/superbol_languageclient.ml
bclement-ocp: Will do, thought I did already.

src/vscode/superbol-vscode-platform/superbol_instance.ml
nberth: We have many places where `"superbol-free"` is hard-coded. Even if it's the default maybe some of those places are not really useful. For instance, it might be worth just not to call `set_bundled_superbol` from `superbol_vscode_platform.ml` if `find_superbol` fails.

bclement-ocp: Removed that `superbol` by storing the instance in a reference.

src/vscode/superbol-vscode-platform/superbol_vscode_platform.ml
```diff
@@ -230,14 +230,17 @@ let activate (extension : Vscode.ExtensionContext.t) =
       "superbol-free"
   in
 
-  Superbol_instance.set_bundled_superbol instance
-    bundled_superbol;
+  let instance = Superbol_instance.make ~bundled_superbol () in
+  current_instance := Some instance;
```
nberth, line 234: May `activate` be called more than once with distinct `ExtensionContext`s? If yes, I guess that shall be handled later when we deal with storage URIs.

nberth (approved):
Ok for me ;-)

src/vscode/superbol-vscode-platform/superbol_vscode_platform.ml
bclement-ocp: No, `activate` is only called once (as is `deactivate`). Well, barring VS Code bugs, but that's the contract (cf https://code.visualstudio.com/api/references/activation-events#Start-up "An extension must export an activate() function from its main module and it will be invoked only once by VS Code when any of the specified activation events is emitted.").

PR #146, opened 2024-01-24
Handle PERFORM FOREVER statements in LSP

engboris:
This statement appears in https://github.com/OCamlPro/gnucobol-contrib/blob/master/samples/DBsample/PostgreSQL/example1/PGTEST1.cob (line 174-208)

```cobol
      *>------------------------------------------------------------------------
       MAIN-PGTEST1 SECTION.
      *>------------------------------------------------------------------------
      
          PERFORM FOREVER
             DISPLAY HEADER-SCREEN END-DISPLAY  
             DISPLAY MAIN-FUNCTION-SCREEN END-DISPLAY
             DISPLAY MESSAGE-SCREEN END-DISPLAY 
             ACCEPT MAIN-FUNCTION-SCREEN END-ACCEPT
      
      *>     init message       
             INITIALIZE WS-MSG
             DISPLAY MESSAGE-SCREEN END-DISPLAY 
             
             EVALUATE TRUE
                WHEN V-FNC-F1
                   PERFORM FNC-CONNECT-SCREEN
      
                WHEN V-FNC-F2
                   PERFORM FNC-DISCONNECT
                   
                WHEN V-FNC-F9
                   EXIT PERFORM
                   
                WHEN OTHER
                   MOVE "Please select a valid function key" 
                     TO WS-MSG-1 OF WS-MSG
             END-EVALUATE
          END-PERFORM
      
          STOP RUN
      
          .
       MAIN-PGTEST1-EX.
          EXIT.
```

According to https://github.com/OCamlPro/gnucobol/blob/gcos4gnucobol-3.x/cobc/reserved.c (line 1406), it comes from the standard COBOL2002 C/S and can occur in the context of a `PERFORM` statement (in a `PERFORM` scope terminating by `END-PERFORM`.


src/lsp/cobol_parser/grammar.mly
```diff
@@ -3384,18 +3384,19 @@ let reversed_or_no_rewind_opt :=
 (* TODO: COB85 first format also takes instructions... *)
 
 %public let unconditional_action := ~ = perform_statement; < >
-let perform_statement :=
+let perform_statement [@context perform_stmt] :=
+  | PERFORM; po = ro(perform_phrase); isl = imp_stmts; END_PERFORM;
```
nberth, line 3388: Please reorder the two rules to have a diff focused on the essential changes: only the line starting with `let perform_statement` appears to need modification.

src/lsp/cobol_ptree/branching_statements.ml
```diff
@@ -99,6 +99,7 @@ and perform_inline_stmt =
   }
 
 and perform_mode =
+  | PerformForever
```
nberth, line 102: Please add this as last constructor as it's not so usual, with a `(* COBOL2002, according to GnuCOBOL's sources *)` comment (this is not even mentioned in COBOL2014, so I'm a bit surprised).

nberth (commented):
Two minor comments, mostly to make a better diff ;-)

src/lsp/cobol_ptree/branching_statements.ml
nberth: Now that I think about it, the comment was attached to the `FOREVER` keyword itself, that appears somewhere else in the grammar.
In this particular case this seems a Micro-focus extension.

nberth: Not even Micro-focus, just COBOL-IT…

nberth: So the comment should just state: `GC extension` (GC standing for GnuCOBOL).

src/lsp/cobol_parser/grammar.mly
engboris: I actually exchanged 
```ocaml
| PERFORM; po = ro(perform_phrase); isl = imp_stmts; END_PERFORM; [...]
```
and 
```ocaml
| PERFORM; i = procedure_name; [...]
```
in purpose in order to prevent `FOREVER` to be parsed as a procedure name in `PERFORM FOREVER`.

src/lsp/cobol_ptree/branching_statements.ml
engboris: Do you mean the comment in GnuCOBOL's sources is wrong? How do you know it is just a COBOL-IT extension?

src/lsp/cobol_parser/grammar.mly
nberth: The sequencing of production rules in such a grammar specification does not impact how a token is parsed: no precedence is derived from this order (apart maybe from cases of conflicts that are resolved arbitrarily — but we don't allow such a case to occur).

src/lsp/cobol_ptree/branching_statements.ml
nberth: I've looked for it in several documents. The only place I could find it is on a reference for COBOL-IT.

My (almost sure) guess is: the `COBOL2002 C/S` comment comes from the introduction of the keyword `FOREVER` in the `RETRY` context. It was later added to the `PERFORM` context without the comment being adjusted.

nberth (approved):
Ok to merge, thanks!

PR #155, opened 2024-01-30
Bugfix: upon rewind, clear source overlay locations cache earlier

nberth:
This avoids a potential infinite loop in cache lookup by the source overlay manager. Before this change, new tokens and location limits were produced before the cache was cleared, which could led to an infinite loop during lookup in the cache.

### DIFF
--- a/src/lsp/cobol_preproc/src_overlay.ml
+++ b/src/lsp/cobol_preproc/src_overlay.ml
@@ -25,7 +25,7 @@ module type MANAGER = sig
   val link_limits: limit -> limit -> unit
   val join_limits: limit * limit -> srcloc
   val dummy_limit: limit
-  val restart_at: limit -> unit
+  val restart: ?at:limit -> unit -> unit
 end
 
 (** Overlay limits (internal) *)
@@ -162,7 +162,7 @@ let join_limits: manager -> limit * limit -> srcloc = fun ctx (s, e) ->
   with Not_found ->
     join_failure (s, e)
 
-let restart_at ctx _limit =
+let restart ?at:_ ctx =
   Links.clear ctx.cache
 
 module New_manager (Id: sig val name: string end) () : MANAGER = struct
@@ -172,5 +172,5 @@ module New_manager (Id: sig val name: string end) () : MANAGER = struct
   let link_limits = link_limits ctx
   let join_limits = join_limits ctx
   let dummy_limit = Lexing.dummy_pos
-  let restart_at = restart_at ctx
+  let restart ?at () = restart ?at ctx
 end
--- a/src/lsp/cobol_preproc/src_overlay.ml
+++ b/src/lsp/cobol_preproc/src_overlay.ml
@@ -87,14 +87,16 @@ let new_manager: string -> manager =
     location; for any given file, must be called with the leftmost location
     first. *)
 let limits: manager -> srcloc -> limit * limit = fun ctx loc ->
-  let s, e = match Cobol_common.Srcloc.as_unique_lexloc loc with
+  let left, right = match Cobol_common.Srcloc.as_unique_lexloc loc with
     | Some lexloc -> lexloc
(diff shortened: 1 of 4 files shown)


lefessan:
I am still seeing the bug.

To reproduce, you might try: take any cobol file, replace a `01 ...` declaration by a `COPY file` with `file` containing the declaration. Then, go back to the initial declaration. Superbol takes 100% of the CPU in an infinite loop in `try_cache_from`.

lefessan:
See #156 for another fix for the same bug.

lefessan:
I couldn't reproduce the bug with this PR.

src/lsp/cobol_preproc/src_overlay.ml
```diff
@@ -91,8 +91,9 @@ let limits: manager -> srcloc -> limit * limit = fun ctx loc ->
     | Some lexloc -> lexloc
     | _ -> Limit.make_virtual (), Limit.make_virtual ()
   in
-  Links.replace ctx.right_of s (loc, e);  (* Replace to deal with duplicates. *)
```
lefessan, line 94: Could u take this opportunity to rename `e` in `right` and `s` in `left`, or some other **long** names ?

src/lsp/cobol_preproc/src_overlay.ml
```diff
@@ -162,7 +163,7 @@ let join_limits: manager -> limit * limit -> srcloc = fun ctx (s, e) ->
   with Not_found ->
     join_failure (s, e)
 
-let restart_at ctx _limit =
+let restart ?at:_ ctx =
```
lefessan, line 166: I think it is not a good idea to have an unused optional argument...

src/lsp/cobol_preproc/src_overlay.ml
```diff
@@ -172,5 +173,5 @@ module New_manager (Id: sig val name: string end) () : MANAGER = struct
   let link_limits = link_limits ctx
   let join_limits = join_limits ctx
   let dummy_limit = Lexing.dummy_pos
-  let restart_at = restart_at ctx
+  let restart ?at () = restart ?at ctx
```
lefessan, line 176: `at` is not used by `restart`, so what's the point of giving it ?

src/lsp/cobol_preproc/src_overlay.mli
```diff
@@ -29,7 +29,12 @@ module type MANAGER = sig
   val id: string
 
   (** [limits loc] creates and returns the left- and right-limit to the given
-      location. *)
+      location.
+
+      Note: if [loc] overlaps with locations that have already been fed to this
+      function, then {!restart} must first be called with a limit that is
+      strictly at the left of [loc] ({i i.e,} that comes earlier in the stream
```
lefessan, line 36: But the `at` argument of `restart` is not used, no ?

lefessan (changes requested):

lefessan:
I think you should prepare a talk at some point on the full behavior of the preprocessor (and maybe even parser).

src/lsp/cobol_preproc/src_overlay.ml
nberth: Cheap future-proofing.
It could be quite possible to optimize things in terms of cache management and representation of the limits. The idea is to have an interface that is not likely to change whenever this optimisation is implemented.

nberth:
> I think you should prepare a talk at some point on the full behavior of the preprocessor (and maybe even parser).

Yes possibly.

src/lsp/cobol_preproc/src_overlay.ml
lefessan: Given that this optimisation is not going to be done very soon, it is better to keep a non misleading type (i.e. where all arguments are actually used). 

When the optimization will be implemented, it will be easy to change the interface to an extra non-optional argument to find all occurrences that need to be modified.

lefessan (approved):

PR #156, opened 2024-01-30
fix infinite loop in src_overlay

lefessan:
This PR contains two commits:
* one to add debug stuff to help solving bugs by outputing in logs and reducing the size of these logs by detecting cycles
* one to fix the bug in src_overlay, that happens because the code is looking for an exact match, that actually does not exist

### DIFF
--- a/src/lsp/cobol_preproc/src_overlay.ml
+++ b/src/lsp/cobol_preproc/src_overlay.ml
@@ -13,6 +13,17 @@
 
 open Cobol_common.Srcloc.TYPES
 
+let debug_oc = ref None
+
+(*
+let message fmt =
+  Printf.kprintf (fun s ->
+      match !debug_oc with
+      | None -> ()
+      | Some oc ->
+        Printf.fprintf oc "OVERLAY: %s\n%!" s) fmt
+*)
+
 module TYPES = struct
   type limit = Lexing.position
 end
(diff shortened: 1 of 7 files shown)


src/lsp/cobol_preproc/src_overlay.ml
```diff
@@ -129,7 +140,13 @@ let join_limits: manager -> limit * limit -> srcloc = fun ctx (s, e) ->
       if e.pos_cnum  = e'.pos_cnum &&   (* compare only the fields that matter *)
          e.pos_fname = e'.pos_fname
       then (Links.replace ctx.cache s (loc, e); loc)
-      else try_cache_from ~loc @@ Links.find ctx.over_right_gap e'
+      else
+      if e.pos_cnum  = e'.pos_cnum &&   (* compare only the fields that matter *)
```
nberth, line 144: Please don't merge this, this is not a proper fix (some `pos_cnum`s are not positive).

nberth: (Not the same code I saw earlier)

lefessan: fixed

src/lsp/cobol_preproc/src_overlay.ml
```diff
@@ -129,7 +140,13 @@ let join_limits: manager -> limit * limit -> srcloc = fun ctx (s, e) ->
       if e.pos_cnum  = e'.pos_cnum &&   (* compare only the fields that matter *)
          e.pos_fname = e'.pos_fname
       then (Links.replace ctx.cache s (loc, e); loc)
-      else try_cache_from ~loc @@ Links.find ctx.over_right_gap e'
+      else
+      if e.pos_cnum  < e'.pos_cnum &&   (* detect that there is no exact match *)
```
nberth, line 144: For a "cleaner" temporary fix (that would make the bug just "less frequent"), please guard this with `e.pos_cnum > 0`.

PR #157, opened 2024-02-01
Refactoring: remove library interfaces

lefessan:
Library interfaces are generally a wrong good idea: it hides modules that could be useful later, it renames modules whose source files are then harder to locate, it may contain code that should be in other modules.

So, we must avoid them.


nberth (changes requested):
As a whole this PR gathers some changes  I adhere to, and some I am strongly opposed to:

- I am against of removing interface module in libraries that mostly define types and associated utilities (like the parse tree), then I do not want to have to type `Types` all the time, while I still want to mention where the types are from in other libraries;
- I am also against  the renaming of many module names for which the prefix indicates where a module belongs to (`Src_` and `Preproc_` in the preprocessor, `data_` in the types for representing the COBO data, etc). When editing files in various libraries at the same time, those prefixes are very useful. Complicated changes in the code often imply editing several libraries at the same time, so I think on the long run that outweighs the additional on-boarding cost.
- I am in favor of removing the `include`s interface modules (so, having `Main` modules that gather principal points).

All in all, I'd rather opt for a shortened version of this PR where the only remaining interface modules would be a list of module bindings that only remove their respective library-specific prefix (like `Lsp_`).

lefessan:
FWIW, the subject is not whether we keep some package interfaces or not, we will get rid of all of them for readability and navigation issues. The question is more how to make it while keeping some consistency and ease of programming and qualifying things.

PR #158, opened 2024-02-02
Make goto-def and find-refs return locations within copybooks

nberth:
Before this change, copied definitions and references were localized on the corresponding `COPY` within the main source program.


engboris (approved):
OK for me (nice feature)

PR #162, opened 2024-02-05
Fix lookup when `COPY <text-name> IN/OF <relative path>` is used

nberth:
Before these changes, a `COPY something OF "."` interpreted "." w.r.t the current working directory.  Now, the "." is interpreted w.r.t the directory name of the file where the `COPY` statement is found.

This allows the LSP to properly handle one of the NIST-COBOL85 tests, whatever the project directory (from which every relative path is interpreted).

### DIFF
--- a/src/lsp/cobol_common/copybook.ml
+++ b/src/lsp/cobol_common/copybook.ml
@@ -13,6 +13,8 @@
 
 open Ez_file.V1
 
+type fileloc = [ `Word of string | `Alphanum of string ]
+
 type lookup_info =
   {
     libname: string;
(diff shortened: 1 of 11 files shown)


engboris (commented):
It works as expected.

However, when I'm in a working directory `A` and open a file `test.cob` outside of `A` in VSCode, the `COPY` statement without `IN "."` does not find the copy file. Is it something expected? Is the default directory always expected to be the current working directory even when editing an external file?

My working environment:
- `cd gnucobol-contrib/samples/worldcities`
- `code .`
- I open some file `test-copy.cob`
```cob
       IDENTIFICATION DIVISION.
       PROGRAM-ID. test-copy.

       DATA DIVISION.
       WORKING-STORAGE SECTION.
       COPY "statement.cpy".

       PROCEDURE DIVISION.
           DISPLAY SPACE.
           STOP RUN.
```

nberth:
> It works as expected.
> 
> However, when I'm in a working directory `A` and open a file `test.cob` outside of `A` in VSCode, the `COPY` statement without `IN "."` does not find the copy file. Is it something expected? Is the default directory always expected to be the current working directory even when editing an external file?
> 
> My working environment:
> 
> * `cd gnucobol-contrib/samples/worldcities`
> * `code .`
> * I open some file `test-copy.cob`
> 
> ```cobol
>        IDENTIFICATION DIVISION.
>        PROGRAM-ID. test-copy.
> 
>        DATA DIVISION.
>        WORKING-STORAGE SECTION.
>        COPY "statement.cpy".
> 
>        PROCEDURE DIVISION.
>            DISPLAY SPACE.
>            STOP RUN.
> ```

For the version with `IN/OF` as dealt with in this PR, the operand on the right indicates a library directory.
For `COPY` without `IN/OF …`, there is a configurable search path; I don't think putting a copybook folder that is outside of the project directory in there will pose many issues; if they need to be opened, then probably an additional directory will be part of the workspace. That will be explained in the doc.

engboris (approved):
OK

PR #163, opened 2024-02-06
Fix source locations of data division's sections

nberth:
Such locations now span over section headers; before, they started at the first item.


engboris (approved):
OK

PR #164, opened 2024-02-06
Substitute `npx yarn` for plain `yarn`

engboris:
The `-g` option in `npm install -g yarn` is necessary in order to be able to use the `yarn install` command.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,13 +84,8 @@ server and the VSCode extension from sources.
 
    [^drom-version]: Current version is 0.9.2~dev3 (commit 63a5770).
 
-3. To build the VSCode extension, you also need to have
-   [node.js](https://nodejs.org/), and install `yarn` via:
-
-   ```bash
-   npm install -g yarn
-   yarn install
-   ```
+3. Install [node.js](https://nodejs.org/) (version >=5.2.0) if it
+   is not already installed.
 
 4. You can then install all remaining dependencies, and compile the
    LSP server along with the VSCode extension:
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ server and the VSCode extension from sources.
    [node.js](https://nodejs.org/), and install `yarn` via:
 
    ```bash
-   npm install yarn
+   npm install -g yarn
    yarn install
    ```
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -88,7 +88,7 @@ server and the VSCode extension from sources.
    [node.js](https://nodejs.org/), and install `yarn` via:
 
    ```bash
-   npm install yarn
+   npm install -g yarn
    yarn install
    ```
```
nberth, line 94: Could you add a little comment line right below that says ("Note this will install `yarn` globally.") ?

(Same in `building-superbol.rst`)

nberth (commented):
Do you know if that's a strong requirement? Maybe some `eval $(npm env)` (or whatever) command is enough afterwards to make the command available in the shell.

engboris:
> Do you know if that's a strong requirement? Maybe some `eval $(npm env)` (or whatever) command is enough afterwards to make the command available in the shell.

It's not a strong requirement. A simple and maybe better way to do it is to simply run the locally installed yarn:
```
npm install yarn
./node_modules/.bin/yarn install
```

And in more recent versions (>=5.2.0), we can simply do
```
npx yarn install
```
which is even more convenient.

nberth:
I am not familiar with JS/node stuff. I'd favor `npx yarn install` if that version is "old-enough" (like one/two year old) ;-)

engboris:
Ok I just remarked that the Makefile uses the `yarn` command so it does not work with a local installation. Should the Makefile be modified as well?

nberth:
> Ok I just remarked that the Makefile uses the `yarn` command so it does not work with a local installation. Should the Makefile be modified as well?

I just tried as well. Indeed, setting `YARN = npx yarn $(YARN_ARGS)` in `Makefile.header` seems to do the job.
Could you then add
```make
.PHONY: node-setup
node-setup:
	npx installl yarn # with possibly an option to disable interactive validation
dev-deps: node-setup
```
somewhere in `Makefile.header` (like right below `clean: clean-execs`)?

nberth (approved):
I think that's fine. Thanks!

PR #154, opened 2024-01-26
Better config UX

nberth:
Included changes mostly include:
- Early loading of project caches upon startup (when the LSP client/VS Code provides a list of workspace directories);
- Use of VS Code's settings to configure projects in the LSP server;
- A new command to save the current VS Code configuration to a `superbol.toml`, for each project directory opened in a workspace. Once this file exists, changes to VS Code's settings are ignored (but trigger a warning to the user).

PR #165, opened 2024-02-06
Manifest improvements

nberth:
Gathers some changes from #154, for easier review:
- Customization of activation events for the extension;
- A JSON schema that can be used by TOML-dedicated LSPs to validate and edit the `superbol.toml` configuration file.

Regarding the latter point, this PR adds a dependency to the `tamasfe.even-better-toml` VS Code extension. The idea may be, if and when we find out how that can be done, to have this as a "suggestion" only, not as a mandatory dependency.

### DIFF
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -10,4 +10,3 @@ _out/
 !CHANGELOG.md
 !LICENSE.md
 !package.json
-!schemas/
--- a/.vscodeignore
+++ b/.vscodeignore
@@ -10,4 +10,4 @@ _out/
 !CHANGELOG.md
 !LICENSE.md
 !package.json
-!astexplorer/dist/index.html
+!schemas/
--- a/package.json
+++ b/package.json
@@ -80,13 +80,15 @@
           "title": "SuperBOL executable",
           "type": "string",
           "default": "",
-          "description": "Name of the `superbol-free` executable if available in PATH; may be an absolute path otherwise. Leave empty to use the bundled `superbol-free`, if available."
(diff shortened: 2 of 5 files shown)


engboris:
Is the purpose of the JSON schema to allow the user to edit the `superbol.toml` configuration file within VSCode's UI?

nberth:
> Is the purpose of the JSON schema to allow the user to edit the `superbol.toml` configuration file within VSCode's UI?

Such a schema is used by extensions like `even-better-toml` to help in editing the relevant TOML file, by performing basic checks on the underlying structure, or by providing auto-completion and the likes.

engboris:
I'm divided for the second feature. I'm not fan of having such a dependency because it should be the choice of the user only. But it is also a good way to ensure a good and controlled user experience. Making it an option / a suggestion would probably be better (as suggested).

Do you know if there are any VSCode extensions with such optional dependencies?

nberth:
> I'm divided for the second feature. I'm not fan of having such a dependency because it should be the choice of the user only. But it is also a good way to ensure a good and controlled user experience. Making it an option / a suggestion would probably be better (as suggested).
> 
> Do you know if there are any VSCode extensions with such optional dependencies?

Alas, I did not find any way to just "suggest" an extension. Plus, at the moment, the idea is to provide an extension that works fine "out-of-the-box", without needs for further tuning.

In addition, I have tested several TOML extensions, and this one was the most complete I could find (also LSP-based, that's probably why).

engboris (approved):
Ok for me. This dependency can be discussed later.

PR #167, opened 2024-02-06
Optionally pass a `--storage-directory` argument to the LSP server

nberth:
This change avoids creating a `_superbol` folder in each project's root directory.  A storage space provided by the client is used instead.

For now, the "global" storage space provided by VS Code for the extension is used. It is unclear (/ I did not search for) who is responsible for cleaning that space (or the local storage).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [0.1.4] Next release
 
 ### Added
+- Configuration flag for caching in storage provided by Visual Studio Code [#167](https://github.com/OCamlPro/superbol-studio-oss/pull/167)
 - Configuration setting for copybook filename extensions [#332](https://github.com/OCamlPro/superbol-studio-oss/pull/332), with updated JSON schema [#333](https://github.com/OCamlPro/superbol-studio-oss/pull/333)
 - COBOL language configuration for highlighting matching brackets and auto-insertion of line numbers in fixed-format code [#330](https://github.com/OCamlPro/superbol-studio-oss/pull/330)
 
--- a/package.json
+++ b/package.json
@@ -128,7 +128,7 @@
             "cbx"
           ],
           "items": "string",
-          "order": 1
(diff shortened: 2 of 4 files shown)


nberth:
Actually, just found out we don't have a "simple" way of dealing with the cleanup problem by only relying on VS Code (*cf* https://github.com/microsoft/vscode/issues/32461).

Temporary solution: do not merge this and continue relying on a `_superbol` sub-folder at the root of the project.

Later: use the extension context's `globalState` to store an association list of *workspace*/*project directories* → *cache storage directories*, and implement a basic cleanup strategy upon activation/deactivation.

engboris:
Can you explain why you use [ExtensionContext.globalStorageUri](https://code.visualstudio.com/api/references/vscode-api#ExtensionContext.globalStorageUri) instead of [ExtensionContext.storageUri](https://code.visualstudio.com/api/references/vscode-api#ExtensionContext.storageUri)? I see you have a comment about it but maybe I'm missing some subtlety.

Do you expect the storage to be accessible from *all* workspaces?

nberth:
> Can you explain why you use [ExtensionContext.globalStorageUri](https://code.visualstudio.com/api/references/vscode-api#ExtensionContext.globalStorageUri) instead of [ExtensionContext.storageUri](https://code.visualstudio.com/api/references/vscode-api#ExtensionContext.storageUri)? I see you have a comment about it but maybe I'm missing some subtlety.
> 
> Do you expect the storage to be accessible from _all_ workspaces?

Yes. We store a cache on a per-project directory basis, not per-workspace. So, it could be relevant to reload a project from the same cache even if that's from two distinct workspaces (especially if the configurations match, which is likely if projects have a `superbol.toml`).

However, as I noted in the previous comment, we have no "simple" way of cleaning caches. And apparently, this problem would also arise if we were to use workspace-storage instead.

emilienlemaire (approved):

PR #168, opened 2024-02-07
Better handle config lookup on Windose

nberth:
… as is typically the case on Windose.

### DIFF
--- a/src/lsp/cobol_config/cobol_config.ml
+++ b/src/lsp/cobol_config/cobol_config.ml
@@ -46,10 +46,14 @@ let default_search_path =
       | Some d -> [ d ]
       | None -> []
     and xdg_conf_dir = match Sys.getenv_opt "XDG_CONFIG_HOME" with
-      | Some p -> p
-      | None -> Sys.getenv "HOME" // ".config"
+      | Some p -> Some p
+      | None -> try Some (Sys.getenv "HOME" // ".config") with Not_found -> None
     in
-    cwd :: xdg_conf_dir // "superbol" :: cob_conf_dir
+    let xdg_superbol_dir = match xdg_conf_dir with
+      | None -> []
+      | Some d -> [d // "superbol"]
+    in
+    cwd :: xdg_superbol_dir @ cob_conf_dir
   end
 
 let retrieve_search_path ?search_path () =
--- a/src/lsp/cobol_config/cobol_config.ml
+++ b/src/lsp/cobol_config/cobol_config.ml
@@ -40,20 +40,26 @@ let print_options: Pretty.delayed = fun ppf ->
     ppf !all_configs
 
 let default_search_path =
+  let append root ~sub = match root with
+    | Some d -> [d // sub]
+    | None -> []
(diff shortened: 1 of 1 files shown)


ddeclerck:
On Windows, you could use the `USERPROFILE` environment variable for this purpose.
But in fact it is more common to use `APPDATA` (usually = `%USERPROFILE%\AppData\Roaming`) and `LOCALAPPDATA` (usually = `%USERPROFILE%\AppData\Local`).
The difference between the two appears when one connects to sevral Windows machine with a profile stored on some central server : anything in `APPDATA` is will always be available no matter the machine, while anything in `LOCALAPPDATA` is machine-specific.

nberth:
Thanks @ddeclerck for the tip! I've amended the search path accordingly. Is it still worth checking for `XDG_*` on such systems as well?

ddeclerck:
> Thanks @ddeclerck for the tip! I've amended the search path accordingly. Is it still worth checking for `XDG_*` on such systems as well?

Well, if one installs X11 under Windows (either through Cygwin or MSYS2), `XDG_*` *might* be defined (though it was not the case when I tried). But since we're running on Windows, we *do* expect things to go in `APPDATA` and `LOCALAPPDATA` anyways (those variables are defined even if you run from a Cygwin or MSYS2 shell).

nberth:
> I'm wondering why you're putting `COB_CONFIG_DIR` in the same list ?

This path is where we want to lookup GnuCOBOL's configuration files; maybe it makes no sense to expect some under those directories…
In any case, that path should be more configurable in the future.

ddeclerck:
So all those paths will only ever be *read* and never written ; right ?

nberth:
> So all those paths will only ever be _read_ and never written ; right ?

No, that's just a search path to read configuration files. There's no plan to write anything in particular in those directories.

src/lsp/cobol_config/cobol_config.ml
```diff
@@ -40,16 +40,26 @@ let print_options: Pretty.delayed = fun ppf ->
...
+    let cob_config_dir = Option.to_list (Sys.getenv_opt "COB_CONFIG_DIR") in
+    let xdg_superbol_dir =  (* can is this available on win32/cygwin as well? *)
+      append ~sub:"superbol" @@
+      match Sys.getenv_opt "XDG_CONFIG_HOME" with
+      | Some p -> Some p
+      | None -> Option.map (fun d -> d // ".config") (Sys.getenv_opt "HOME")
+    in
+    let system_specific =
+      if Sys.(win32 || cygwin)
+      then append ~sub:"SuperBOL" (Sys.getenv_opt "APPDATA") @
```
ddeclerck, line 58: Does this compile ? I'd expect matching the result of `getenv_opt` with `Some`/`None` before calling `append` ?

ddeclerck: Ah, forget that, I had not read the code of `append`, which does the match.

ddeclerck (approved):

Issue #142, opened 2024-01-22
Add `check syntax` sub-command for `superbol-free`


nberth:

GitMensch:
What would this be used for? Isn't the LSP running in watch mode while editing a file, or would this subcommand raise a message where the"normal" LSP ignores something or doesn't understand?

nberth:
This command shall be a sub-command of the `superbol-free` CLI executable, that will allow us to better report missing support for compiler directives or COBOL constructs in the LSP's grammar.

PR #169, opened 2024-02-08
Add check syntax command

engboris:
See issue https://github.com/OCamlPro/superbol-studio-oss/issues/142

This PR adds a "check syntax" command to check the presence of syntax errors in a COBOL file. It can be used as follows:
```
./superbol-free[...] check syntax <path to COBOL file>
```

### DIFF
--- a/src/lsp/superbol_free_lib/command_check_syntax.ml
+++ b/src/lsp/superbol_free_lib/command_check_syntax.ml
@@ -32,7 +32,7 @@ let cmd =
     let common, common_args = Common_args.get () in
 
   EZCMD.sub
-    "parse file"            (* add the corresponding line in Main.subcommands *)
+    "check syntax"            (* add the corresponding line in Main.subcommands *)
     begin fun () ->
       action (common ()) !files |>
       Seq.iter Cobol_common.Diagnostics.sink_result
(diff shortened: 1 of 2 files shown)


src/lsp/superbol_free_lib/command_check_syntax.ml
```diff
@@ -0,0 +1,51 @@
...
+  in
+  List.to_seq files |>
+  Seq.map begin fun filename ->
+    Pretty.out "@[Considering@ `%s'@]@." filename;
+    Cobol_preproc.Input.from ~filename ~f:parse
+  end
+
+let cmd =
+  let files = ref [] in
+    let common, common_args = Common_args.get () in
```
nberth, line 32: There's a weird indentation here.

src/lsp/superbol_free_lib/command_check_syntax.ml
```diff
@@ -0,0 +1,51 @@
...
+  let parse input =
+    input |>
+    Cobol_preproc.preprocessor ~options:preproc_options |>
+    Cobol_parser.parse_simple ~options:parser_options
+  in
+  List.to_seq files |>
+  Seq.map begin fun filename ->
+    Pretty.out "@[Considering@ `%s'@]@." filename;
+    Cobol_preproc.Input.from ~filename ~f:parse
+  end
```
nberth, line 28: With this, if multiple input file names are provided, we'd get a bunch of `Considering…`, and then all diagnostics gathered for every file.

Instead, directly using `List.iter` and then calling `Cobol_common.Diagnostics.sink_result` here would show diagnostics directly underneath the `Considering...` message. That would probably be better for a CLI.

nberth (approved):
Thanks!
Just a little suggestion before we merge.

nberth (approved):
Ok to merge as soon as the CI terminates ;-)

PR #172, opened 2024-02-09
Consider Linux-specific standard GnuCOBOL configuration paths

nberth:
This should allow the extension to work properly on typical Linux setups.

### DIFF
--- a/src/lsp/cobol_config/cobol_config.ml
+++ b/src/lsp/cobol_config/cobol_config.ml
@@ -57,7 +57,8 @@ let default_search_path =
       if Sys.(win32 || cygwin)
       then append ~sub:"SuperBOL" (Sys.getenv_opt "APPDATA") @
            append ~sub:"SuperBOL" (Sys.getenv_opt "LOCALAPPDATA")
-      else []
+      else ["/usr/local/share/gnucobol/config";
+            "/usr/share/gnucobol/config"]
     in
     cwd ::  cob_config_dir @ xdg_superbol_dir @ system_specific
   end
--- a/src/lsp/cobol_config/cobol_config.ml
+++ b/src/lsp/cobol_config/cobol_config.ml
@@ -52,15 +52,18 @@ let default_search_path =
       match Sys.getenv_opt "XDG_CONFIG_HOME" with
       | Some p -> Some p
       | None -> Option.map (fun d -> d // ".config") (Sys.getenv_opt "HOME")
-    in
-    let system_specific =
+    and windose_specific =
       if Sys.(win32 || cygwin)
       then append ~sub:"SuperBOL" (Sys.getenv_opt "APPDATA") @
            append ~sub:"SuperBOL" (Sys.getenv_opt "LOCALAPPDATA")
-      else ["/usr/local/share/gnucobol/config";
+      else []
+    and unix_specific =
+      if Sys.(unix || cygwin)
+      then ["/usr/local/share/gnucobol/config";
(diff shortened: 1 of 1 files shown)


src/lsp/cobol_config/cobol_config.ml
```diff
@@ -57,7 +57,8 @@ let default_search_path =
       if Sys.(win32 || cygwin)
       then append ~sub:"SuperBOL" (Sys.getenv_opt "APPDATA") @
            append ~sub:"SuperBOL" (Sys.getenv_opt "LOCALAPPDATA")
-      else []
+      else ["/usr/local/share/gnucobol/config";
+            "/usr/share/gnucobol/config"]
     in
```
ddeclerck, line 62: Actually on Cygwin you *might* want to have both `*APPDATA` and `*/share/gnucobol/config` (you don't know if the user installed a "Cygwin" GnuCOBOL or a native GnuCobol).

ddeclerck: It also depends what kind of superbol is installed though. A Windows-native superbol has no idea about Unix paths, while a Cygwin superbol can handle both Unix and Windows paths...

ddeclerck: Not even mentionning MSYS2, which builds native binaries (can't be distinguished with `Sys.*`) that can use both Windows-style and Unix-style paths.

nberth: Hum… do you think it is worth hacking with a check using `EzFile.FileOS.dir_separator = "/"`?

ddeclerck: I have no idea how `EzFile.FileOS.dir_separator` behaves when run from a Cygwin or MSYS2 shell.

I ran a few tests to clarify the behavior of Cygwin and MSYS2 with respect to paths. Turns out I was partly wrong.
Here's what I learnt :
- native Windows binaries built by Cygwin or MSYS2 have no idea about Unix paths
- "Cygwin" binaries built by Cygwin understand both Windows and Unix paths
- MSYS2 can also produce "Cygwin" binaries, though people don't do that (the whole point of MSYS2 is to build native binaries)

So basically, when you have `Sys.cygwin = true`, then you can add both `*APPDATA` and `*/share/gnucobol/config` to your search path (even if it seems unlikely people will ever run a Cygwin superbol server).

nberth: Ok I've amended to take cygwin into considerations.

ddeclerck: Perfect !

src/lsp/cobol_config/cobol_config.ml
```diff
@@ -52,14 +52,18 @@ let default_search_path =
       match Sys.getenv_opt "XDG_CONFIG_HOME" with
       | Some p -> Some p
       | None -> Option.map (fun d -> d // ".config") (Sys.getenv_opt "HOME")
-    in
-    let system_specific =
+    and windose_specific =
```
ddeclerck, line 55: You do realize it's spelled "Windows", right ? ;)

nberth: Yes. I'll let you fix that later if that really makes you sad :stuck_out_tongue_winking_eye:

ddeclerck: I'll survive ;)

ddeclerck (approved):

PR #174, opened 2024-02-26
Synchronize debug extension

nberth:


src/lsp/superbol_free_lib/vscode_extension.ml
```diff
@@ -211,54 +225,6 @@ let contributes =
                         }
                 }
         } |})
-        ~configurationSnippets: [
-          Manifest.any {| {
-                                "label": "SuperBOL: debug (launch)",
-                                "description": "New SuperBOL launch request",
```
ddeclerck, line 217: What's sthe reason behind this deletion ?

nberth: Just found out the extension itself seems to provide the snippets (via some code, I don't remember where)

nberth: (I'm testing it again top triple-check)

nberth: (`extension.ts`, `provideDebugConfigurations`)

ddeclerck: If I remember correctly, the snippets in `package.json` and the configurations provided by code are used in different contextes (I think one provides the defaults when you try to run but don't have a `launch.json` while the other one is indeed used to generate the configurations in `launch.json`).

ddeclerck: I had noted  this :
* when launch.json absent:
  => use package.json/initialConfigurations AND/OR extension.ts/provideDebugConfigurations
* when launch.json present:
  => use package.json/configurationSnippets

nberth: Hum… actually the only purpose of these snippets seems to be for intellisense, when editing the `launch.json`. They do not seem to be used to provide the defaults, but to "document" the defaults, as I understand it. In my case, the new `gdbtty` field was there when VSCode auto-filled the `launch.json`, even though I hadn't edited the snippet yet; this is why I assumed it was not really useful.

I'll remove that last commit and add the `gdbtty` to the snippet instead; overall, it's quite annoying that most of these infos need to be duplicated.

ddeclerck (approved):

PR #177, opened 2024-02-28
Add CHAINING keyword in grammar

engboris:
I encountered it in `gnucobol-contrib/samples/worldcities/worldcities0.cbl`.

The `CHAINING` keyword behaves the same as the `USING` keyword which is already handled by the parser. As far as I understand, the difference is that `USING` is associated to a `CALL` statement, whereas `CHAINING` is associated to a `CHAIN` statement (see [Microfocus documentation](https://www.microfocus.com/documentation/cobol-it/4-11/user-guide/compiler-suite/procedure-division-clause.html)). It seems to be an extension, according to GnuCOBOL's source code.

## References

- [Microfocus documentation](https://www.microfocus.com/documentation/cobol-it/4-11/user-guide/compiler-suite/procedure-division-clause.html)
- [GnuCOBOL parser.y](https://github.com/OCamlPro/gnucobol/blob/57133577c7e5524f14cf3cb868df3c1fdbc21ad7/cobc/parser.y#L10962)
- [GnuCOBOL reserved.c](https://github.com/OCamlPro/gnucobol/blob/57133577c7e5524f14cf3cb868df3c1fdbc21ad7/cobc/reserved.c#L622)


nberth (commented):
Looks fine grammar-wise. However, we may need to register in the AST what keyword is used between `CHAINING` and `USING`.

engboris:
I don't see how it could be useful. Do you have an idea?

nberth:
> I don't see how it could be useful. Do you have an idea?

I see in the documentation that "the `CHAIN` statement and `CHAINING`
clause are recognized, and syntax is validated. However, the `CHAIN`
statement/`CHAINING` clause are otherwise treated as commentary."  In
addition, we may later want to check that targets of `CHAIN` are indeed
declared with `CHAINING` instead of `USING`.

src/lsp/cobol_ptree/proc_division_visitor.ml
```diff
@@ -132,12 +135,26 @@ let fold_raising_phrase (v: _ #folder) =
 let fold_raising_phrase' (v: _ #folder) =
   handle' v#fold_raising_phrase' ~fold:fold_raising_phrase v
 
+let fold_procedure_calling_style (v: _ #folder) =
+  leaf v#fold_procedure_calling_style
+
+let fold_procedure_calling_style' (v: _ #folder) =
+  leaf' v#fold_procedure_calling_style'
```
nberth, line 142: I don't think we need `fold_procedure_calling_style'` (neither this binding, nor the method in `folder`).

In any case, that one is wrong (I'm not sure why I left the `leaf'` helper as it is misleading), as in this case `fold_procedure_calling_style` would not be called.

src/lsp/cobol_ptree/proc_division_visitor.ml
```diff
@@ -132,12 +135,26 @@ let fold_raising_phrase (v: _ #folder) =
...
+
+let fold_procedure_args (v: _ #folder) =
+  handle v#fold_procedure_args
+    ~continue:begin fun
+      {procedure_calling_style=style; procedure_by_clause=args} x -> x
```
nberth, line 147: Please follow the overall style, in particular for visitors: with the spaces after/before `{` and `}`, and around `=` (and no line exceeding 80 chars appox.):
```ocaml
    ~continue:begin fun { procedure_calling_style = style;
                          procedure_by_clause = args } x -> x 
   …
```

nberth (commented):
Just minor changes and it's good to merge.

nberth (approved):
Perfect! Good to merge :tada:

PR #180, opened 2024-03-04
Terminating visitors

nberth:
Additionally, attach some more locations to the AST.


engboris:
It seems to me that some comments are not recognized anymore: stars `*` on 7th column for free format (when not followed by a `>` symbol). The problems occurs on several files of `gnucobol-contrib` (typically the last worldcities files).

nberth:
> It seems to me that some comments are not recognized anymore: stars `*` on 7th column for free format (when not followed by a `>` symbol). The problem occurs on several files of `gnucobol-contrib` (typically the last worldcities files).

Single stars anywhere in non-fixed format do not denote beginning comment, so that's normal.

nberth:
> Should all these files be considered wrong, then?

This PR does not involve the pre-processor, so I doubt there is an issue with comments to be discussed here. Please open an dedicated issue with more details ;-)

PR #186, opened 2024-03-05
Rely on debug configurations provided by the debugger extension

nberth:
This allows users to run programs (with or without a debugging session) with default configurations when a `launch.json` is not present.

Also restrict version of `jsoo_promise` to work around a breaking change in its API.


nberth:
(converting to dratf while I investigate what's causing issues in vscode js stubs (sic))

PR #189, opened 2024-03-05
Fix missing terminating newline for TOML files

### LINKED ISSUE #152: Lost ability to properly parse TOML files that do not end with a newline character
In these cases we get a `Ez_toml.Types.Error (_, 0, 0)` exception.

lefessan:
fix #152

### DIFF
--- a/src/vendor/ez_toml/v1.ml
+++ b/src/vendor/ez_toml/v1.ml
@@ -38,7 +38,17 @@ module TOML = struct
           lexbuf.lex_curr_p <- {lexbuf.lex_curr_p with pos_fname = file}
     end;
     try
-      let lines = Internal_parser.toml Internal_lexer.tomlex lexbuf in
+      let eof = ref false in
+      let tomlex lexbuf =
+        if !eof then Internal_parser.EOF else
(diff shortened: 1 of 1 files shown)


src/vendor/ez_toml/v1.ml
```diff
@@ -38,7 +38,17 @@ module TOML = struct
           lexbuf.lex_curr_p <- {lexbuf.lex_curr_p with pos_fname = file}
     end;
     try
-      let lines = Internal_parser.toml Internal_lexer.tomlex lexbuf in
+      let eof = ref false in
```
nberth, line 41: I though the problem was a missing newline expected by the parser. Is this actually fixing the issue?

lefessan: The idea is to always send a EOL token before the EOF token, so there is always an EOL at the end...

lefessan: And yes, I added a test in the original `ez_toml` and the problem is fixed there.

nberth: Ah sure! I originally did not distinguish `EOF` and `EOL`.

nberth (approved):

Issue #197, opened 2024-03-05
bug: words are "recognized" with the default "C style"


GitMensch:
While the COBOL word recognition was added to the debugging part (`registerEvaluatableExpressionProvider`) it is missing in the plain editor. I don't know if this is about the tm syntax or an lsp part, but if I put the cursor into `SOME-VAR`, then the complete word should be highlighted, not only the part before/after the hyphen.

nberth:
I think this may boil down to customizing word separators in the text editor. Apparently this can be done via the extension (*cf* https://github.com/Microsoft/vscode/issues/33701).

PR #200, opened 2024-03-07
Use syntax file for colorization of text that miss semantic tokens

nberth:
Also fix issue #197.


GitMensch:
While you've likely "checked manually" - can this at least be partially tested with automated tests?

That still may be moved to a new issue, but should be considered.

nberth:
I think testing should be moved to a new issue, as right now I have no idea how to test (rendering/editing) behaviors of VSCode.

engboris (approved):

PR #201, opened 2024-03-07
command 'superbol-free reformat to-free FILES.cob'

lefessan:
Translate files from fixed to free format

In case it can be useful to somebody else...


GitMensch:
friendly ping @nberth @lefessan

PR #206, opened 2024-03-08
SuperBOL Studio Pack

nberth:
- Update the main extension's description, and remove its dependency to `tamasfe.even-better-toml`;
- Add an extension pack into the `studio` sub-directory. Calling `vsce package --out superbol-studio-pack-0.1.0.vsix` into `studio` will build a pack that can directly be published on the marketplace.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,6 +252,13 @@ control.  Extensions dedicated to the edition of TOML files, such as
 provide the same level of assistance as when you edit
 `.vscode/settings.json`.
 
+> [!TIP]
+>
+> Install the
+> [`OCamlPro.SuperBOL-studio-pack`](https://marketplace.visualstudio.com/items?itemName=OCamlPro.SuperBOL-studio-pack)
+> extension to get SuperBOL Studio OSS and `tamasfe.even-better-toml`
+> altogether.
+
 ![Editing `superbol.toml`](./assets/superbol-editing-superbol.toml.png)
 
 ### Further documentation
--- a/package.json
+++ b/package.json
@@ -446,9 +446,9 @@
   ],
   "name": "SuperBOL",
   "displayName": "SuperBOL Studio OSS",
-  "description": "Provides a COBOL mode in VSCode, based on the SuperBOL LSP server for COBOL",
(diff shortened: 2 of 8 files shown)


engboris (approved):

GitMensch:
@nberth  *Not* requesting, but wondering: As it says "optional source level debugger" - should that be in a separate extension, too, instead of being included?
I _think_ both the debug extension and the lsp one operate individually, no?

nberth:
> @nberth _Not_ requesting, but wondering: As it says "optional source level debugger" - should that be in a separate extension, too, instead of being included? I _think_ both the debug extension and the lsp one operate individually, no?

I'm sure of one dependency: our version of the debug extension depends on the "main" extension. The reverse case needs to be checked further.
At the moment I'm rather in favor of removing that "optional" (an oversight on my part for that PR).

nberth:
> > I'm sure of one dependency: our version of the debug extension depends on the "main" extension.
> 
> Can you elaborate further, please?

The debug extension requires build tasks that are defined in the main/platform extension since they share configuration settings (and the LSP server itself plays a role in dealing with project/workspace-specific settings).

> Even then it seems useful to have those published separate with the right dependency set (allows updates separately, allows users that only want the editing part [for example because they don't use GnuCOBOL yet] to get only one), ....

I agree this brings greater flexibility. On the other hand, another aspect we want to consider is user-friendliness: having only one extension to install and deal with is often much simpler for average users.

GitMensch:
vscode will install the cobol one with the debugger one, and if they follow the tip and install the extension pack, then they have all three ;-)

nberth:
> vscode will install the cobol one with the debugger one, and if they follow the tip and install the extension pack, then they have all three ;-)

But vscode won't install the debugger one with the COBOL LSP one, and then having the debugging extension depends on users "following the tip" (which is not at all a given in most cases). What would be VSCode's behavior if a user just wants to start debugging then?
I assume developers that are familiar with how VSCode extensions typically operate will follow the tip. What concerns me is… all the others.

GitMensch:
As noted *no request*  just a note. I'm fine with whatever you choose.

PR #210, opened 2024-03-11
Support dynamic addition/removal of workspace folders

nberth:


engboris:
How can I test it? I'm not sure to understand what "dynamic addition/removal of workspace folders" means

nberth:
> How can I test it? I'm not sure to understand what "dynamic addition/removal of workspace folders" means

You may test in VSCode by opening a folder with some COBOL in VSCode, then doing some workspace folder manipulations like adding/removing folders from the workspace (and, notably, observing that diagnostics are kept between operations; to do that you'll need to setup folder projects with COBOL85 dialect until #213 is merged).

engboris (approved):

PR #223, opened 2024-03-20
Fix missing comments in superbol.toml (fix in ez_toml)

lefessan:
Also small fix in `superbol-free project init`


nberth (approved):
Fine!

PR #227, opened 2024-03-20
Handle arguments of intrinsic TRIM function

engboris:
# Problem

The `TRIM` intrinsic function takes either `LEADING` or `TRAILING` (which are also reserved keywords using for string functions for instance) as arguments. However, the grammar of the LSP does not recognise those arguments.

:warning: This pull request is not completed. It looks like `FUNCTION TRIM(..., LEADING)` even after my changes is not recognised so I am probably missing something.

### DIFF
--- a/src/lsp/cobol_config/reserved_words.ml
+++ b/src/lsp/cobol_config/reserved_words.ml
@@ -906,6 +906,7 @@ let reserved_words =
     "TRANSFORM";
     "TRANSPARENT";
     "TREE-VIEW";
+    "TRIM";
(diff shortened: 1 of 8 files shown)


nberth:
@emilienlemaire What is your intention with the latest commit? This PR is already a bit outdated, so I'd suggest a rebase on `master` first.

emilienlemaire:
I just wanted to see what was done before going further, and checking why it doesn't work, so I needed the tests so it's easier to check whether it works or not

PR #229, opened 2024-03-20
Support for END PROGRAM without a name

lefessan:


test/output-tests/run_extensions.expected
```diff
@@ -1724,7 +1724,7 @@ run_extensions.at-2143-prog.cob:8.18:
 ----                     ^
```
nberth, line 1724: Seems like we may need to support alphanumeric program IDs.

src/lsp/cobol_ptree/compilation_group.ml
```diff
@@ -28,7 +28,7 @@ type program_unit =
     program_env: environment_division with_loc option;
     program_data: data_division with_loc option;
     program_proc: procedure_division with_loc option;
-    program_end_name: name with_loc option
+    program_end_name: ident_or_nonnum with_loc option
```
nberth, line 31: I think this should be `name_or_alphanum` instead.

nberth (changes requested):
We should probably be more restrictive in terms of what can come after `END PROGRAM`. `ident` is somewhat misleading by modern standards: it actually encompasses many things like function calls and pointers…

src/lsp/cobol_parser/grammar.mly
```diff
@@ -813,14 +820,21 @@ let input_output_section :=
...
+    { { file_control_paragraph = None;
+        io_control_paragraph = Some io_control_paragraph } }
 
 
 (* - ENVIRONMENT DIVISION / INPUT-OUTPUT SECTION / FILE-CONTROL PARAGRAPH -- *)
 
 let file_control_paragraph :=
- | FILE_CONTROL; "."; ~ = rl(select); < > (* COB85: non-empty list *)
+ | FILE_CONTROL; "." ; ~ = rl(select); < > (* COB85: non-empty list *)
+ | s = select; sl = rl(select); { s :: sl }
```
nberth, line 837: This non-empty list can be coded with `| ~ = rnel(select); < >

src/lsp/cobol_parser/grammar.mly
```diff
@@ -1046,8 +1067,15 @@ let section(K, L) ==
 let section_header_n_items(K, L) ==
   | K; SECTION; "."; ~ = rl(loc(L)); < >
 
+(* Section with optional header, especially on MF *)
+let section_opt_header(K, L) ==
+  | ~ = loc(section_opt_header_n_items(K,L)); < >
+let section_opt_header_n_items(K, L) ==
+  | K; SECTION; "."; ~ = rl(loc(L)); < >
+  | e = loc(L); el = rl(loc(L)); { e :: el }
```
nberth, line 1075: `| ~ = rnel(loc(L)); < >`

src/lsp/cobol_ptree/compilation_group.ml
lefessan: I have tried, but I gave up. I cannot even understand how to translate an `ident_or_aphanum` into a `string` in `terms.ml`.

nberth: What do you mean? Here we need `name_or_alphanum`, not `ident_or_alphanum`. Why do you need a `string`?

lefessan: Because later in the analysis phases, some functions want to get a string for the program name, so I needed to do some conversions. But the Terms format and how to access it remains a complete mystery for me...

src/lsp/cobol_parser/text_tokenizer.ml
```diff
@@ -234,6 +234,7 @@ let preproc_n_combine_tokens ~source_format =
     | [PROGRAM_ID]                   -> Error `MissingInputs
     | PROGRAM_ID :: PERIOD :: _        -> info_word_after 2
     | PROGRAM_ID :: _                  -> info_word_after 1
+    | END :: PROGRAM :: _              -> info_word_after 2
```
nberth, line 237:
```suggestion

    | [END] | [END; PROGRAM]          -> Error `MissingInputs
    | END :: PROGRAM :: _              -> info_word_after 2
```

src/lsp/cobol_parser/text_tokenizer.ml
```diff
@@ -234,6 +234,7 @@ let preproc_n_combine_tokens ~source_format =
     | [PROGRAM_ID]                   -> Error `MissingInputs
```
nberth, line 234:
```suggestion
    | [PROGRAM_ID]
    | [PROGRAM_ID; PERIOD]           -> Error `MissingInputs
```

nberth (commented):
Just two fixes.

src/lsp/cobol_parser/text_tokenizer.ml
nberth:
```suggestion
    | [PROGRAM_ID]
    | [PROGRAM_ID; PERIOD]           -> Error `MissingInputs
```

nberth:
```suggestion
    | [PROGRAM_ID]
    | [PROGRAM_ID; PERIOD]           -> Error `MissingInputs
```

PR #235, opened 2024-03-21
Further support for MF dialect

lefessan:
- Add support for `ACCEPT ... FROM {ARGUMENT/ENVIRONMENT}-VALUE`;
- Add `GOBACK RETURNING`;
- Relax order and optional headers in `DATA DIVISION`;
- New testsuite for `autofonce`, to be updated with cases found in the wild.


nberth (dismissed):
Ok!

nberth (approved):

PR #237, opened 2024-03-21
Add GIVING synonym to CALL's RETURNING

ddeclerck:


nberth (approved):
Fine.
We might need to register what keyword was used for dialect-checks, but that's left for *later*.

PR #239, opened 2024-03-21
Fix ALTERNATE clause of SELECT statement

engboris:
I followed [GnuCOBOL's](https://github.com/OCamlPro/gnucobol/blob/89c45a3bc80c2ca6302382bd40fdc585436d65f9/cobc/parser.y#L5635) syntax where not only `KEY` but also `RECORD` is optional.

Such syntax is used in [BankDemo](https://github.com/MicroFocus/BankDemo/blob/c1d49ed2fc9285d6da3b25f01e8265aa86747847/sources/cobol/core/ZBNKLOAD.cbl#L77).

### DIFF
--- a/src/lsp/cobol_parser/grammar.mly
+++ b/src/lsp/cobol_parser/grammar.mly
@@ -896,7 +896,7 @@ let access_mode :=
  | SEQUENTIAL; {AccessModeSequential}
 
 let alternate_record_key_clause :=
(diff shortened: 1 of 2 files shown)


nberth:
I think you need to rebase this one first; some of the changes have already been merged.

nberth (approved):
Thanks!

PR #241, opened 2024-03-22
Allow empty statement list in PERFORM

ddeclerck:


nberth (approved):
Too simple to be true… or is it? :wink:

ddeclerck:
> Too simple to be true… or is it? 😉

Yes indeed. I think the mandatory `END-PERFORM` in this format helps.

PR #242, opened 2024-03-22
Support `EXEC`/`END-EXEC` blocks (within `PROCEDURE DIVISION` for now)

### LINKED ISSUE #161: handling of conditional compilation (and `EXEC SQL`)
This sample program was shared on the IBM extension:

```cobol
       IDENTIFICATION DIVISION.
       PROGRAM-ID. TSTSQLCA.
       DATA DIVISION.
       WORKING-STORAGE SECTION.
       01  HOWMUCH PIC S9(9) COMP-5.
       >>define isDB2 as b'1'
       >>if isDB2
           exec sql include sqlca end-exec.
       >>end-if
       PROCEDURE DIVISION.
       MAIN.
           exec sql
             select count(*) into :HOWMUCH
             from SYSIBM.SYSDUMM1
           end-exec
           if SQLCODE not = 0 and SQLCODE not = +100
             display "Oups !"
           end-if
           .
       BYE.
           goback.
```

I'm "quite sure" that the LSP also has a problem here (it should error if the `>>define` line is removed, and pass otherwise (and of course, know that `HOMMUCH` is used within the SQL statement).

nberth:
Closes #161 (once extended to the `DATA DIVISION`)


ddeclerck (approved):

GitMensch:
Are test cases missing in this PR or did I just not spot them?

In any case: ignoring those in cobc with outputting one error per XYZ type (from EXEC XYZ) would be very nice. That would have to be done in the preparser (not including the block but a #line directive with the end position instead). Could you give that a go?

nberth:
> Are test cases missing in this PR or did I just not spot them?

No indeed, there are no test cases. Support is temporary, as for the moment the type of EXEC (*e.g,* `SQL`, `CICS`) is ignored, and all EXEC/END-EXEC blocks are treated in the same way. Some tests may come in a subsequent PR (I'll do a dedicated issue for this), but that could wait until we have more support for one of these languages.

> In any case: ignoring those in cobc with outputting one error per XYZ type (from EXEC XYZ) would be very nice. That would have to be done in the preparser (not including the block but a #line directive with the end position instead). Could you give that a go?

That's would be an interesting addition indeed. I may have a go at this sometime soon.

PR #245, opened 2024-03-27
More MF support in DISPLAY/ACCEPT statements

ddeclerck:


src/lsp/cobol_parser/grammar.mly
```diff
@@ -2894,15 +2913,33 @@ let accept_misc :=
...
+ | ~ = at_position;                                  <AcceptAt>
+ | FROM; CRT;                                        {AcceptFromCRT}   (* MF *)
+ | MODE; IS?; BLOCK;                                 {AcceptModeBlock} (* MF *)
+ | WITH?; ~ = nell(loc(accept_with_clause)); %prec lowest <AcceptWith> (* MF *)
+   (* Note: GnuCOBOL makes "WITH" optional *)
 
-let accept_with_clause :=
- | UPDATE;                      {AcceptUpdate}
+let accept_with_clause [@context accept_stmt] := (* MF *)
  | ~ = screen_attribute_clause; <AcceptAttribute>
```
ddeclerck, line 2923: If I don't add the `accept_stmt` context here, some context-sensitive keywords in this clause are not recognized (eg: AUTO)

src/lsp/cobol_parser/grammar_tokens.mly
```diff
@@ -106,7 +106,7 @@
 %token ATTRIBUTE                    [@keyword]                [@contexts set_stmt]
 %token ATTRIBUTES                   [@keyword]                [@contexts ]
 %token AUTHOR                       [@keyword]
-%token AUTO                         [@keyword]                [@contexts screen_descr_entry]
+%token AUTO                         [@keyword "AUTO", "AUTO-SKIP"] [@contexts screen_descr_entry, accept_stmt]
```
nberth, line 109: `AUTO-SKIP` should not be included here: it's typically reserved on a dialect-specific basis (*e.g,* `mf.words`).

Same for the tokens below (`BEEP`, `FULL`, etc).

src/lsp/cobol_ptree/branching_statements.ml
```diff
@@ -208,9 +208,26 @@ and accept_clause =
   | AcceptWith of accept_with_clause with_loc list                      (* MF *)
 
 and accept_with_clause =
-  | AcceptUpdate
   | AcceptAttribute of Data_descr.screen_attribute_clause
-
+  | AcceptAuto
```
nberth, line 212: I suggest we take this opportunity to start sorting those constructors alphanumerically.

src/lsp/cobol_ptree/branching_statements.ml
```diff
@@ -248,10 +265,14 @@ and display_device_mnemonic =
  | DisplayDeviceCommandLine
 
 and display_with_clause =
-  | DisplayBlank of Data_descr.blank_clause
-  | DisplayErase of Data_descr.erase_clause
   | DisplayAttribute of Data_descr.screen_attribute_clause
-
+  | DisplayGrid
```
nberth, line 269: Same as `accept_with_clause`

src/lsp/cobol_ptree/statements_visitor.ml
```diff
@@ -829,8 +829,27 @@ and fold_accept_clause' (v: _ #folder) : accept_clause with_loc -> 'a -> 'a =
...
+      | AcceptAuto -> x
+      | AcceptFull -> x
+      | AcceptRequired -> x
+      | AcceptSecure -> x
+      | AcceptGrid -> x
+      | AcceptLeftLine -> x
+      | AcceptOverLine -> x
+      | AcceptSize ii -> fold_ident_or_intlit v ii x
+      | AcceptControl i -> fold_ident v i x
+      | AcceptPromptCharacter (inno) ->
```
nberth, line 842: No need for the parentheses around `inno`

test/output-tests/syn_misc.expected
```diff
@@ -223,7 +223,7 @@ syn_misc.at-1706-prog.cob:12.24:
 ----                           ^
   13              ACCEPT X WITH NO UPDATE  END-ACCEPT.
   14              ACCEPT X WITH NO DEFAULT END-ACCEPT.
->> Hint: Missing UPDATE
+>> Hint: Missing AUTO
```
nberth, line 226: Are we still missing the `WITH DEFAULT`…?

src/lsp/cobol_parser/grammar.mly
```diff
@@ -2894,15 +2913,33 @@ let accept_misc :=
...
- | WITH; ~ = loc(accept_with_clause)+; <AcceptWith>      (* MF *)
-
-let accept_with_clause :=
- | UPDATE;                      {AcceptUpdate}
- | ~ = screen_attribute_clause; <AcceptAttribute>
-(* TODO: MF allows many clauses *)
+  | ~ = at_position;                                  <AcceptAt>
+  | FROM; CRT;                                        {AcceptFromCRT}   (* MF *)
+  | MODE; IS?; BLOCK;                                 {AcceptModeBlock} (* MF *)
+  | WITH?; ~ = nell(loc(accept_with_clause)); %prec lowest <AcceptWith> (* MF *)
```
nberth, line 2919: `rnell` instead of `nell` may avoid unnecessary suggestions upon syntax errors.

src/lsp/cobol_parser/grammar.mly
```diff
@@ -2894,15 +2913,33 @@ let accept_misc :=
...
+    (* Note: GnuCOBOL makes "WITH" optional *)
+
+let accept_with_clause := (* MF *)
```
nberth, line 2922: Here you could do:
```ocaml
let accept_with_clause [@symbol "<accept-with-clause>"] :=
```
to get better syntax hints.

Same for `display_with_clause`.

nberth (changes requested):
Some things might need reconsidering.

src/lsp/cobol_ptree/branching_statements.ml
ddeclerck: Actually I sorted them "according to the 2023 standard", and extension as they appear in the MF docs.

test/output-tests/syn_misc.expected
ddeclerck: Yeah, this one is neither COBOL 2023 nor Microfocus.

src/lsp/cobol_parser/grammar.mly
ddeclerck: How does that work ?

src/lsp/cobol_ptree/statements_visitor.ml
ddeclerck: I know, but I like them :p

test/output-tests/syn_misc.expected
ddeclerck: Seems like it's an ACUCOBOL stuff. Though ACUCOBOL does not have the negative forms (`NO DEFAULT` , `NO UODATE`)

src/lsp/cobol_ptree/branching_statements.ml
nberth: Ok, I did not notice that.

src/lsp/cobol_parser/grammar_tokens.mly
nberth: As a general rule, anything that's not in `reserved.c` from GnuCOBOL should not appear here.

src/lsp/cobol_ptree/statements_visitor.ml
nberth: Actually I think they have a different semantics depending on the constructor: `type t = A of (int * string)` is not equivalent to `type t' = A of int * string`, and I see the additional pair of parentheses as similarly misleading.

test/output-tests/syn_misc.expected
nberth: Can be left for later. Can you try to see if the `[@symbol …]` suggestion fixes the hint?

src/lsp/cobol_parser/grammar.mly
nberth: When an `accept_with_clause` is actually required as per the grammar, the recovery mechanism uses that `<accept-with-clause>` string instead of a possibly arbitrary sequence of tokens that forms a valid `accept_with_clause`. The auto-completed/missing sequence of tokens may be selected less arbitrarily by using `[@cost]` annotations, but adding the suggested string makes the sequence more readable.

test/output-tests/syn_misc.expected
ddeclerck: That results in this diff :
```
->> Hint: Missing AUTO
+>> Hint: Missing rnell_rev_loc_accept_with_clause__
```
Pretty weird, ain't it ?

src/lsp/cobol_parser/grammar.mly
nberth: @engboris I think that aspect is interesting for you as well (I planned to explained that last week, but somehow I lacked the opportunity to do so).

src/lsp/cobol_ptree/statements_visitor.ml
ddeclerck: It's true when defining the constructor, but when using a constructor, it does not matter - hopefully the type helps disambiguate (otherwise, how could you tell the type by solely looking at `A (42, "toto")` ?)

test/output-tests/syn_misc.expected
nberth: Hum… maybe the `rnell` is messing things up. What do you get with `nell` that you used originally instead.

src/lsp/cobol_ptree/statements_visitor.ml
nberth: Sure the meaning is the same when matching, the difference is in type definitions.
In any case Boris and I voted the same, for not using unneeded parentheses whenever possible in such patterns.

test/output-tests/syn_misc.expected
ddeclerck: With `nell` i just get `>> Hint: Missing AUTO`.

src/lsp/cobol_ptree/statements_visitor.ml
ddeclerck: That has been changed already. I wonder why this is not showing as outdated.

test/output-tests/syn_misc.expected
ddeclerck: Done. Works better (still need to use `nell` instead of `rnell` BTW).

nberth: Ok thanks for the notice. Maybe `rnell` always messes up hints (I need to check that) (ideally we'd need the hint/symbol as a parameter to `rnell`… but I guess that's not covered by menhir ;-)

nberth (approved):

PR #246, opened 2024-03-27
Add specialized AST nodes for scalars

nberth:


ddeclerck (approved):
This is tricky indeed.
Didn't see anything suspicious.

engboris:
Why is it called "scalar"? And can you confirm that it corresponds to GnuCOBOL's `x`?

PR #247, opened 2024-03-27
Add GO TO ENTRY statement

engboris:
The `GO TO ENTRY` and `GO TO ENTRY ... DEPENDING` statements were not recognised.

DRAFT: currently stuck on some "production never reduced" errors. There are obvious conflicts but I did not looked enough into it: `GO TO ENTRY ...` can be parsed as

```
GO TO
ENTRY ...
```

or

```
GO TO ENTRY ...
```

### DIFF
--- a/src/lsp/cobol_ptree/branching_statements.ml
+++ b/src/lsp/cobol_ptree/branching_statements.ml
@@ -31,6 +31,16 @@ type goto_stmt =
 let pp_goto_stmt ppf { goto_target } =
   Fmt.pf ppf "@[GO TO @[%a@]" pp_procedure_name' goto_target
 
+(* GOTO ENTRY *)
+type goto_entry =
+  {
+    goto_entry_target: alphanum_string
+  }
+[@@deriving ord]
+
+let pp_goto_entry ppf { goto_entry_target } =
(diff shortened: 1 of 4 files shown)


src/lsp/cobol_ptree/branching_statements.ml
```diff
@@ -31,6 +31,16 @@ type goto_stmt =
 let pp_goto_stmt ppf { goto_target } =
   Fmt.pf ppf "@[GO TO @[%a@]" pp_procedure_name' goto_target
 
+(* GOTO ENTRY *)
+type goto_entry =
```
nberth, line 35: One last thing: `goto_entry_stmt` instead of just `goto_entry` (and appropriately renamed in visitors).

nberth: Actually, I didn't pay enough attention before, but I don't see any point in having four distinct constructors for `GoTo*`. At worse we should have `GoTo` and `GoToEntry`, but the `DEPENDING` part is not really relevant.
The distinctions should come in `goto_stmt`, with dedicated constructors and inline records.

nberth (approved):
I guess we can merge as is, and I'll file an issue to merge the `GoTo` constructors.

PR #248, opened 2024-03-27
Fix parsing of Z-literals

ddeclerck:


src/lsp/cobol_parser/text_lexer.ml
```diff
@@ -257,7 +257,7 @@ let decode_symbolic_ebcdics' ~quotation w =
       end
   in
   ACC.result ~diags
-    (Grammar_tokens.ALPHANUM { str; quotation; hexadecimal = false } &@<- w)
+    (Grammar_tokens.ALPHANUM { str; quotation; hexadecimal = false; runtime_repr = Native_bytes } &@<- w)
```
nberth, line 260: Please could you cut the line right before `runtime_repr`?

src/lsp/cobol_parser/text_tokenizer.ml
```diff
@@ -380,13 +380,14 @@ let tokens_of_word { persist = { lexer; _ }; _ }
   : text_word with_loc -> tokens * Parser_diagnostics.t =
   fun { payload = c; loc } ->
   let tok t = [t &@ loc], Parser_diagnostics.none in
-  let alphanum ~hexadecimal str qte =
+  let alphanum ~hexadecimal ~repr str qte =
```
nberth, line 383: Considering this is quite local, you should just `?(repr = Native_bytes)` instead.

src/lsp/cobol_ptree/terms.ml
```diff
@@ -60,15 +60,27 @@ type alphanum_quote =
...
   {
     str: string;
     quotation: alphanum_quote;
     hexadecimal: bool;
+    runtime_repr: alphanum_repr;
   }
 [@@deriving ord]
 
-let pp_alphanum_string ppf { hexadecimal; quotation; str } =
+let pp_alphanum_repr ppf repr =
```
nberth, line 77: Not sure doing nothing in some/most cases is what we'd want: I think that could become confusing. I'd go with local decisions on the string representation: one inside `pp_alphanum_string` below, and the other one (even if that's similar) at the other place where `pp_alphanum_repr` is used.

nberth (commented):
Almost ready ;-)

src/lsp/cobol_parser/text_tokenizer.ml
ddeclerck: Maybe. I don't like optional arguments much though (it's easy to forget about them).

nberth (changes requested):
One little detail still :sweat_smile:

src/lsp/cobol_parser/text_tokenizer.ml
```diff
@@ -410,8 +415,8 @@ let tokens_of_word { persist = { lexer; _ }; _ }
     -> boollit ~base:`Hex str
   | Alphanum { knd = Hex; str; qte }
     -> tok @@ ALPHANUM (alphanum ~hexadecimal:true str qte)
-  | Alphanum { knd = NullTerm; str; _ }
-    -> tok @@ NULLIT str
+  | Alphanum { knd = NullTerm; str; qte }
+    -> tok @@ ALPHANUM (alphanum ~hexadecimal:true ~repr:Null_terminated_bytes str qte)
```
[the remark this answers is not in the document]
ddeclerck: Ouch, that was a bad copy-paste ;)

PR #249, opened 2024-03-28
Add USAGE type-name as per MF

ddeclerck:
I was hoping to allow `USAGE type-name` with an optional `USAGE` keyword (to fix some more MF tests), yet having a data description clause starting with a `WORD` token causes way too many conflicts (many clauses contain optional `WORD` tokens...).

EDIT: it's funny that the MF doc doesn't seem to make `USAGE` optional, yet the MF Bank Demo program relies on this...


nberth:
> I was hoping to allow `USAGE type-name` with an optional `USAGE` keyword (to fix some more MF tests), yet having a data description clause starting with a `WORD` token causes way too many conflicts (many clauses contain optional `WORD` tokens...).

I guess the cases where `USAGE` is optional refer to entries that contain an `IS? TYPEDEF` clause,  right. That'll need to be disambiguated after the AST's been constructed, indeed.

ddeclerck:
> > I was hoping to allow `USAGE type-name` with an optional `USAGE` keyword (to fix some more MF tests), yet having a data description clause starting with a `WORD` token causes way too many conflicts (many clauses contain optional `WORD` tokens...).
> 
> I guess the cases where `USAGE` is optional refer to entries that contain an `IS? TYPEDEF` clause, right. That'll need to be disambiguated after the AST's been constructed, indeed.

I meant the `USAGE` keyword is optional when introducing a USAGE clause.

ddeclerck:
> Just missing a `make test-promote` I think.

Yup

ddeclerck:
> > I meant the `USAGE` keyword is optional when introducing a USAGE clause.
> 
> Ah, I had something else in mind. So what you mean is we can have something like `77 <data-item-name> <type-name>.` (we don't parse this yet, do we?).

Yup, we don't parse that, though it is present in some tests.

src/lsp/cobol_parser/grammar.mly
```diff
@@ -1631,6 +1631,10 @@ let data_type_clause := TYPE; TO?; ~ = name; < >
 let usage_clause :=
```
nberth, line 1631:
```suggestion
let usage_clause [@recovery Binary] [@symbol "<usage-clause>"] :=
```
For later: we need to add a `dummy_usage_clause` somewhere in `Cobol_ptree.Dummies` (so we could check later that some nodes have been automatically "guessed").

nberth: Those annotations are sometimes hard to put in the right place. I'll push a commit that moves them to `usage`.

nberth (approved):

PR #254, opened 2024-03-29
Fix optional IS in RECORDING MODE

engboris:
The keyword `IS` can be optional in `RECORDING MODE`.

This syntax occurs in `ACAS/common/acas-get-params.cbl:105`. For instance:

```cobol
 FD  ACAS-Params
     recording mode variable.
```


test/output-tests/reparse.expected
```diff
@@ -79,7 +79,7 @@ Re-parsing `test/testsuite/ibm/ibmmainframes.com/cob12.cbl':
 Re-parsing `test/testsuite/ibm/ibmmainframes.com/cob13.cbl':
  Parse: OK. Reparse: OK.
 Re-parsing `test/testsuite/ibm/ibmmainframes.com/cob14.cbl':
- Parse: Failure.
+ Parse: OK. Reparse: OK.
```
nberth, line 82: Seems like we had a hint for this optional keyword here already ;-)

nberth (approved):

PR #256, opened 2024-03-29
Merge constructors for GO TO statement

### LINKED ISSUE #255: Merge `GoTo*` constructors
There is no point in having four distinct constructors for `GoTo*` in statements. At worse we should have `GoTo` and `GoToEntry`, but the `DEPENDING` part is not really relevant.
The distinctions should come in `goto_stmt`, with dedicated constructors and inline records.

_Originally posted by @nberth in https://github.com/OCamlPro/superbol-studio-oss/pull/247#discussion_r1544488165_

engboris:
Fixes #255

### DIFF
--- a/src/lsp/cobol_parser/grammar.mly
+++ b/src/lsp/cobol_parser/grammar.mly
@@ -4000,15 +4000,16 @@ let start_position ==
 
 %public let unconditional_action := ~ = stop_statement; <Stop>
 let stop_statement :=
- | STOP; ~ = stop_body; < >
-let stop_body [@context stop_stmt] :=
-  | ~ = o(stop_with_arg); <StopArg> (* RM/COBOL extension *)
+  | STOP; { StopArg None }                  (* so `stop_body` is not nullable *)
+  | STOP; ~ = stop_body; < >
+let stop_body [@context stop_stmt] := (* with context: should not accept empty *)
+  | a = stop_with_arg; { StopArg (Some a) }             (* RM/COBOL extension *)
   | RUN; ~ = o(stop_run_returning_body); <StopRun>
-  | ERROR; { StopError }                         (* GCOS *)
+  | ERROR; { StopError }                                              (* GCOS *)
   | THREAD; ~ = o(qualident); <StopThread>
 
 let stop_run_returning_body :=
-  | ~ = scalar; <StopReturningScalar> 
+  | ~ = scalar; <StopReturningScalar>
   | or_(GIVING, RETURNING); o(ADDRESS; OF); ~ = qualident;
     <StopReturningAddress>
   | or_(GIVING, RETURNING); v = integer;
@@ -4020,7 +4021,7 @@ let stop_run_returning_body :=
   | ~ = with_status; <StopReturningStatus>
 
 let stop_with_arg :=
-  | ~ = qualident; <StopWithQualIdent>      (* ~COB85, -COB2002 *)                  
+  | ~ = qualident; <StopWithQualIdent>      (* ~COB85, -COB2002 *)
   | ZERO; { StopWithLiteral (NumFig Zero) }
   | SPACE; { StopWithLiteral (Fig Space) }
   | QUOTE; { StopWithLiteral (Fig Quote) }
--- a/src/lsp/cobol_parser/grammar.mly
+++ b/src/lsp/cobol_parser/grammar.mly
@@ -3356,15 +3356,14 @@ let generate_statement :=
 
 %public let unconditional_action := ~ = go_to_statement; < >
 let go_to_statement :=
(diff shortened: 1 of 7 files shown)


src/lsp/cobol_ptree/branching_statements.ml
```diff
@@ -21,52 +21,38 @@ open Simple_statements
...
-(* GOTO ENTRY DEPENDING *)
-type goto_entry_depending_stmt =
-  {
-    goto_entry_depending_targets: alphanum_string nel;
-    goto_entry_depending_on: ident;
-  }
-[@@deriving ord]
+let pp_goto_stmt ppf = function
+  | GoToSimple { goto_targets; goto_depending_on = None } ->
+    Fmt.pf ppf "@[GO TO @[%a@]"
```
nberth, line 40: There's a missing `@]` at the end of every format string that starts with `@[GO TO`

src/lsp/cobol_ptree/terms.ml
```diff
@@ -97,6 +97,8 @@ let pp_alphanum_string ppf { hexadecimal; quotation; str; runtime_repr } =
   | Simple_quote -> Fmt.pf ppf "'%s'" str
   | Double_quote -> Fmt.pf ppf "\"%s\"" str
 
+let pp_alphanum_string' ppf { payload; _ } = pp_alphanum_string ppf payload
```
nberth, line 100: I think you should use `(pp_with_loc pp_alphanum_string)` instead of that new `pp_alphanum_string'`.

src/lsp/cobol_ptree/branching_statements.ml
```diff
@@ -21,52 +21,38 @@ open Simple_statements
...
-    goto_depending_on: ident;
-  }
+  | GoToSimple of
+    {
+      goto_targets: procedure_name with_loc nel;
+      goto_depending_on: ident option;
+    }
+  | GoToEntry of
+    {
+      goto_entry_targets: alphanum_string with_loc nel;
```
nberth, line 33: That should not be `alphanum_string`; please use `alphanum` instead.

src/lsp/cobol_parser/grammar.mly
```diff
@@ -3354,18 +3354,14 @@ let generate_statement :=
...
    { LoneGoTo }	   (* COB85; obsolete; should be sole statement of paragraph *)
- | GO; TO?; ENTRY; t = alphanum;
-   { GoToEntry { goto_entry_target = t } }
- | GO; TO?; ENTRY; il = nel_(alphanum); DEPENDING; ON?; i = ident;
-   { GoToEntryDepending { goto_entry_depending_targets = il;
-                          goto_entry_depending_on = i } }
+ | GO; TO?; il = nel_(procedure_name); i = o(DEPENDING; ON?; ident);
+   { GoTo (GoToSimple { goto_targets = il;
+                        goto_depending_on = i; } ) }
+ | GO; TO?; ENTRY; il = nel_(loc(alphanum)); i = o(DEPENDING; ON?; ident);
```
nberth, line 3362: You can use a new rule like:
```ocaml
let alphanum_literal == ~ = alphanum; <Alphanum>
```
and then `nel_(loc(alphanum_literal));` here instead.

nberth (commented):
Let's take this opportunity for further cleanup.

We should not use `alphanum_string` directly anywhere in the AST.

src/lsp/cobol_ptree/branching_statements.ml
engboris: The reason I used `alphanum_string` is because no comparison function for `alphanum` is created. I'm not sure how I should deal with that

src/lsp/cobol_ptree/terms.ml
engboris: I remark that there're a lot of `pp'` functions. Should we only use `pp_with_loc` instead? There're a few inconsistencies and I wasn't sure it was on purpose or something to be changed.
If those functions are unwanted, it may be wiser to delete them to prevent other contributors to use it by copying already existing code as I did.

src/lsp/cobol_ptree/branching_statements.ml
nberth: Yes you can do it like this.

src/lsp/cobol_ptree/terms.ml
nberth: The problem is not necessarily the definition of `pp_alphanum_string'`, it is that it does not use `pp_with_loc`.
Since it was used just once I suggested to avoid the redefinition, but the main issue was that `alphanum_string` was used in the first place.
I've had further thoughts about that, so here you can commit whatever works and I'll adjust later.

Issue #153, opened 2024-01-26
Compiler directives for debug


engboris:
In the [GnuCOBOL doc](https://get-superbol.com/gnucobol/manual/chapter2.html#debug-switches), a `-fdebugging-line` option is defined for debugging. The `D` indicator (in the 7th column) is recognised by the LSP but not the `>>D` directive (which exists both in fixed and free format but according to [GnuCOBOL sources](https://github.com/OCamlPro/gnucobol/blob/gcos4gnucobol-3.x/cobc/pplex.l#L259) it is strictly limited to COB2002 for free format). Such directives appears in the [GnuCOBOL test suite](https://github.com/OCamlPro/gnucobol/blob/c0d64addfd83baff21f089e0ab7c92609e291442/tests/testsuite.src/run_fundamental.at#L5107).

I see that in the [preprocessor's parsing of compiler directives](https://github.com/OCamlPro/superbol-studio-oss/blob/master/src/lsp/cobol_preproc/src_lexer.mll#L329), we parse expressions beginning by `>>`, followed by an optional space and finishing by some words used for compiler directives. However, `>>D` is different as it cannot have a space character in-between. For the fixed format, I thought about adding a case:
```ocaml
and fixed_cdir_line marker state          (* `>>`-prefixed compiler directive *)
  = parse
  | 'D'
      {
        fixed_debug_line state lexbuf
      }
  | ' '? cdir_word_suffix
      {
        Src_lexing.cdir_word ~ktkd:gobble_line ~knom:fixed_nominal
          ~marker (Src_lexing.flush_continued state) lexbuf
      }
  | epsilon
      {
        newline_or_eof (Src_lexing.flush_continued state) lexbuf
      }
```
But it would probably be a too ad-hoc fix.

GitMensch:
Ignore what GC does (it may need to be fixed haven't checked) and parse `>>` directives per standard:

> A  compiler  directive  is  composed  of  the  compiler  directive  indicator,  **optionally  followed  by  the  COBOL character space**, followed by compiler-instruction. The compiler directive indicator shall be treated as though it were followed by a space if no space is specified after the indicator.

-> that space is optional

Also, just in case:

>  a compiler directive may be followed only by space characters and an optional inline comment

you may or may not have a testcase for that already...

GitMensch:
[...]

PR #257, opened 2024-04-02
Support more compiler directives

nberth:
Adds support for `>>DEFINE` and simple (cascading) conditional compilation directives. `>>SET` directives are parsed but not interpreted yet. `>>EVALUATE`/`>>WHEN` shall be added in a subsequent PR.


emilienlemaire:
Would this handle #153 ?

nberth:
> Would this handle #153 ?

No, the `>>D` is a floating indicator, not a compiler directive per se. Introducing that one may need quite some care (and we don't have that many examples of its usage I think), so it's left for later.

nberth:
> We would like to be able to parse the whole https://github.com/OCamlPro/gnucobol-contrib/tree/1618cb70d3855fd56a17bb2522c421e38993fc6d/samples/worldcities folder, (with the fixes from [OCamlPro/gnucobol-contrib#1](https://github.com/OCamlPro/gnucobol-contrib/pull/1)), and this is used in some files

I see in there it's not used in floating position… so maybe Boris's suggestion could work (as a temporary measure).

GitMensch:
> No, the >>D is a floating indicator, not a compiler directive per se. Introducing that one may need quite some care (and we don't have that many examples of its usage I think), so it's left for later.

I _think_ only GnuCOBOL supports that and the features is removed in COBOL 2014, so handling this specific one is likely not important.

test/output-tests/run_misc.expected
```diff
@@ -607,139 +607,18 @@ run_misc.at-3603-prog2.cob:9.41:
...
-   8              IF "a" NOT < "1"
-   9 >       >>END-IF
-----         ^^^^^^^^
   10                 DISPLAY "ERROR" END-DISPLAY
-  11              END-IF.
->> Error: Unexpected >>END-IF compiler directive
+  11 >            END-IF.
+----              ^^^^^^
+  12              STOP RUN.
+>> Error: Invalid syntax
```
emilienlemaire, line 617: Is this expected?

emilienlemaire (commented):
Seems all good to me, but one question concerning a test failure

emilienlemaire (approved):
All good for me 👌

GitMensch:
> `>>EVALUATE`/`>>WHEN` shall be added in a subsequent PR.

I'd love a PR for GnuCOBOL on that as well :-)

Concerning this PR: Do you want to do a follow-up for the MF conditional compilation?

```cobol
    78  my-define VALUE 5.
      $IF my-define
          DISPLAY 'something'
      $END
```

nberth:
> Concerning this PR: Do you want to do a follow-up for the MF conditional compilation?
> 
> ```cobol
>     78  my-define VALUE 5.
>       $IF my-define
>           DISPLAY 'something'
>       $END
> ```

The `$`-variants could be handled in this PR.
The handling of 78-level entries on the pre-processor side will come in another PR, though.

src/lsp/cobol_preproc/compdir_grammar.mly
```diff
@@ -30,10 +30,11 @@
...
 
 %token CDIR_DEFINE     [@keyword ">>DEFINE", "$DEFINE"]
-%token CDIR_ELIF       [@keyword ">>ELIF", ">>ELSE-IF"]
-%token CDIR_ELSE       [@keyword ">>ELSE"]
-%token CDIR_END_IF     [@keyword ">>END-IF"]
-%token CDIR_IF         [@keyword ">>IF"]
+%token CDIR_ELIF       [@keyword ">>ELIF", "$ELIF", ">>ELSE-IF", "$ELSE-IF"]
+%token CDIR_ELSE       [@keyword ">>ELSE", "$ELSE"]
+%token CDIR_END        [@keyword "$END"]       (* Note: no `>>END` equivalent *)
+%token CDIR_END_IF     [@keyword ">>END-IF", "$END-IF"]
```
GitMensch, line 36: Note: `$END-IF` _does_ work and is found in existing code-bases, but is undocumented from MF (it is documented as `$END`. You may want to note that so that any later adjustment does not remove that.

src/lsp/cobol_preproc/compdir_grammar.mly
```diff
@@ -30,10 +30,11 @@
...
+%token CDIR_IF         [@keyword ">>IF", "$IF"]
 %token CDIR_SET        [@keyword ">>SET", "$SET"]
 %token CDIR_SOURCE     [@keyword ">>SOURCE", "$SOURCE"]
```
GitMensch, line 39: I _think_ you mean [`$SOURCEFORMAT`](https://www.microfocus.com/documentation/visual-cobol/vc90/DevHub/HRCDRHCDIR63.html). right? Note that this is additional complicated as MF allows optional hyphens, so you also will see `$SOURCE-FORMAT`.

Note for `>>SOURCE` (I don't know if this also applies to MF's version, I'd just assume that) this is "local only" - not sure if that's handled correctly:

>  If a `SOURCE FORMAT` directive is specified in library text, the specified format shall be in effect until another `SOURCE FORMAT` directive is encountered or the end of the library text is reached. When the processing of that library text is completed, the reference format shall revert to the reference format that was in effect for the `COPY` statement that resulted in processing of that library text.

src/lsp/cobol_preproc/compdir_grammar.mly
```diff
@@ -30,10 +30,11 @@
 %token NE              "<>"         [@keyword (* symbol *) "<>"]
 
 %token CDIR_DEFINE     [@keyword ">>DEFINE", "$DEFINE"]
```
GitMensch, line 32: There is no `$DEFINE` - see [MF's `>> DEFINE` docs](https://www.microfocus.com/documentation/visual-cobol/vc90/DevHub/GUID-57C303C9-6446-4054-B2DE-2C2A911BCA9B.html).

Instead they have `level 78` and [`$CONSTANT` directive](https://www.microfocus.com/documentation/visual-cobol/vc90/DevHub/HRCDRHCDIR24.html).

src/lsp/cobol_preproc/compdir_grammar.mly
```diff
@@ -30,10 +30,11 @@
...
-%token CDIR_ELIF       [@keyword ">>ELIF", ">>ELSE-IF"]
-%token CDIR_ELSE       [@keyword ">>ELSE"]
-%token CDIR_END_IF     [@keyword ">>END-IF"]
-%token CDIR_IF         [@keyword ">>IF"]
+%token CDIR_ELIF       [@keyword ">>ELIF", "$ELIF", ">>ELSE-IF", "$ELSE-IF"]
```
GitMensch, line 33: Just a note: There is no `$ELIF` or `$ELSE-IF` in MF, see [their conditional compilation docs](https://www.microfocus.com/documentation/visual-cobol/vc90/DevHub/HRLHLHCOMP0J.html).
There is no `>>ELIF` or `>>ELSE-IF` in the standard (that would be a GC extension).

GitMensch (changes requested):
There are some issues on the MF side (see comments), but I'm mostly concerned about ISO.
I don't know if there's a preparser applied or if the compdir_grammar.mly keywords are taken "verbatim".

ISO says:

> A compiler directive is composed of the compiler directive indicator `>>`, optionally followed by the COBOL character space, followed by compiler-instruction. The compiler directive indicator shall be treated as though it were followed by a space if no space is specified after the indicator.

Also note that "the COBOL character space" can consist of multiple character spaces.

I do wonder: does


```cobol
>> IF something DEFINED
    PERFORM something
  >>  ELSE
    PERFORM other-stuff
   >>END-IF
```
works.

src/lsp/cobol_preproc/compdir_grammar.mly
nberth: Thanks for the notice; I'll add a comment on that.

src/lsp/cobol_preproc/compdir_grammar.mly
nberth: Yes I think I coded the handling of `>>SOURCE` via `COPY` in the way the standards describe it.

About `$SOURCEFORMAT`, I could not see it in GnuCOBOL's preprocessor code, or its tests. I could find `$SET SOURCEFORMAT` though, which should already be handled. (I've just made a little tweak to handle `SOURCE-FORMAT` as well).

src/lsp/cobol_preproc/compdir_grammar.mly
nberth: Thanks for the notice as well.

nberth:
> There are some issues on the MF side (see comments), but I'm mostly concerned about ISO. I don't know if there's a preparser applied or if the compdir_grammar.mly keywords are taken "verbatim".

Those `>>` and `$`-prefixed keywords are recognized during a lexing phase that occurs before, and interveaning spaces are removed at that stage.

> ISO says:
> 
> > A compiler directive is composed of the compiler directive indicator `>>`, optionally followed by the COBOL character space, followed by compiler-instruction. The compiler directive indicator shall be treated as though it were followed by a space if no space is specified after the indicator.

> Also note that "the COBOL character space" can consist of multiple character spaces.
> 
> I do wonder: does
> 
> ```cobol
> >> IF something DEFINED
>     PERFORM something
>   >>  ELSE
>     PERFORM other-stuff
>    >>END-IF
> ```
> 
> works.

For that, I originally followed GnuCOBOL. The version I used did not allow more than one character, so `>>  ELSE` might not be recognized. I think that's an easy PR — but maybe that was fixed recently.

src/lsp/cobol_preproc/compdir_grammar.mly
GitMensch: > About `$SOURCEFORMAT`, I could not see it in GnuCOBOL's preprocessor code, or its tests. I could find `$SET SOURCEFORMAT` though

That was a mistake on my side, everything but the ISO ones (that are in a different part in the MF docs and of course only partial supported) all directives would be either from the command line with `-C`, specified in a directive file,  or in the code as `$SET`. So all is fine here.

GitMensch:
> For that, I originally followed GnuCOBOL. The version I used did not allow more than one character, so `>> ELSE` might not be recognized. I think that's an easy PR — but maybe that was fixed recently.

Oh, please check that then.

Concerning this PR it looks good to me now.

PR #260, opened 2024-04-04
Add syntax tests to the CI

emilienlemaire:


Makefile.header
```diff
@@ -243,3 +243,12 @@ emacs/lsp-superbol-customs.el: $(LSP_MODE_SRCDIR) package.json
 
 test-promote:
```
nberth, line 244: I think we also miss `.PHONY: test-promote`. Could you add this as well?

nberth (commented):
It'd be great to allow tests to pass without `autofonce` installed, as well as a simple way to get it. So I suggest to add:
- `opam install autofonce` as a standalone command in `dev-deps` (in `Makefile.drom-tpl`);
- use `\! opam exec -- autofonce 2>/dev/null ||` as a prefix for the two uses added by this PR.

emilienlemaire:
But `Makefile.drom-tpl` is generated by `drom project`, I can add it as a dependency for the tests

nberth:
> But `Makefile.drom-tpl` is generated by `drom project`, I can add it as a dependency for the tests

Actually the dependency is the reverse: `Makefile` is generated by `drom project` from `Makefile.drom-tpl`.

emilienlemaire:
Oh ok, so the comment `# Generated by "drom project"` at the 2nd line is very miss-leading :sweat_smile:

Makefile.header
```diff
@@ -241,14 +241,15 @@ emacs/lsp-superbol-customs.el: $(LSP_MODE_SRCDIR) package.json
 # --eval "(princ (lsp-generate-settings \"package.json\" 'lsp-superbol))" \
 # --eval '(princ "\n")' \
 
+.PHONY: test-promote
 test-promote:
```
nberth, line 245:
```suggestion
test-promote: test-syntax-promote
```

nberth:
> Oh ok, so the comment `# Generated by "drom project"` at the 2nd line is very miss-leading 😅

Yes… that should be added by `drom` itself, but it doesn't @lefessan (unless I'm mistaken).

nberth:
One check that `syntax` tests are run by the CI, and this PR is good to go ;-)

nberth:
Hum… the workflow file seems to duplicate a lot of what's already in the Makefile, and the new tests are not performed.
To fix that I suggest:
- `cp .github/workflows/workflow.yml{,.drom-tpl}`;
- Edit that `.drom-tpl` file:
  + replace the action in `install test dependencies` by `make dev-deps`;
  + replace the action in `run test suite` by `make test`.

I don't necessarily like that kind of ad hoc fix, but here it's the simplest way I see to fix it.

nberth:
Ouch my bad :facepalm:! It's [this file](https://github.com/OCamlPro/drom-share/blob/master/projects/virtual/dot_github/workflows/workflow.yml) that you need to put in place of `workflow.yml.drom-tpl` and edit.

nberth:
So one of the changes actually need to be in `drom.toml`.

emilienlemaire:
Drom does not seem to update the workflow.yml file

emilienlemaire:
Apparently running `drom dev-deps` update the `workflow.yml` file

nberth:
> Apparently running `drom dev-deps` update the `workflow.yml` file

In general `opam exec -- drom project` is your friend. Watch out for files that are skipped, that we want to avoid.

Makefile.drom-tpl
```diff
@@ -81,9 +81,10 @@ uninstall:
 	opam uninstall .
 
 dev-deps:
+	opam install autofonce
 	opam install ./opam/*.opam ./test/opam/*.opam --deps-only --with-doc --with-test
 
-test:
+test: test-syntax
```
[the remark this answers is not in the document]
nberth: And, possibly `superbol-lsp-server` must be a dependence of `test-syntax`.

nberth:
I think 519610e and the dependency `test-syntax: superbol-lsp-server` are the key changes to make the tests work on the CI.

nberth:
> The `$SUPERBOL` redefinition also is, but I'd like to keep the stderr checks in `syntax-todo.at` (and maybe add them in `syntax-done.at`) as well as the printing of the failed test log so that when the CI fails, we know why

Ok, I'll review the whole after a rebase to get a clean history.

nberth (commented):
I'm ok with most of these changes. Please make sure to:
- perform syntax test only after the main testsuite has completed;
- check that `opam exec -- drom project` does not skip any file.

nberth (approved):
Many thanks!
(I think I'll squash the commits as many of them silently revert each other's changes)

PR #261, opened 2024-04-04
Simplify singleton types in `Cobol_ptree.Terms`

nberth:


ddeclerck (approved):
Makes sense. Much clearer now.

PR #262, opened 2024-04-05
Add intrinsic function handling

emilienlemaire:
TODO:
- [ ] merge tokens `FUNCTION <INTRINSIC>` as `<INTRISINSIC>_FUNC`
- [ ] handle special case for argument (ex: `TRIM (... TRAILING)`)
- [ ] rebase
- [ ] ...

DONE:
- [x] Handle `FUNCTION ... INTRINSIC` in `REPOSITORY`
- [x] Add intrinsic functions as some kind of keyword


emilienlemaire:
This PR will probably replace #227

src/lsp/cobol_parser/grammar.mly
```diff
@@ -2203,16 +2204,15 @@ let subscripts [@recovery []] [@symbol "<subscripts>"] [@cost 0] :=
  | "("; s = subscript_first; ")";                                 { [s] }
 
 (* Only for functions which name is also a keyword in COBOL *)
-let intrinsic_function_name :=
- | LENGTH;      { "LENGTH" }
- | RANDOM;      { "RANDOM" }
- | REVERSE;     { "REVERSE" }
- | SIGN;        { "SIGN" }
- | SUM;         { "SUM" }
+let intrinsic_function_name [@recover dummy_name] [@symbol "<intrinsic function name>"] :=
```
nberth, line 2207: Ideally we wouldn't need that conversion back to strings. Is there a way to avoid it?

src/lsp/cobol_parser/grammar_tokens.mly
```diff
@@ -1399,4 +1399,23 @@
 %token XML_NTEXT                    [@keyword.silenced]
 %token XML_TEXT                     [@keyword.silenced]
 
+(* INTRINSIC FUNCTIONS *)
+%token <string> INTRINSIC_FUNC
```
nberth, line 1403: Here I'd go with a `%token <Cobol_ptree.intrinsic_function_identifier> INTRINSICS_FUNC`.

src/lsp/cobol_ptree/terms.ml
```diff
@@ -269,10 +269,42 @@ and class_ =
...
 and inline_call =                               (* in ancient terms: funident *)
-  {
-    call_fun: name with_loc;
-    call_args: effective_arg list;
-  }
+  | CallFunc of { (*name to avoid clash with call statement *)
+      call_fun: name with_loc;
+      call_args: effective_arg list;
+    }
+  | CallTrim of call_trim_args
```
nberth, line 276: Here I strongly suggest to use inline records when the types are used only once. For the exported types, please add a relevant prefix to the labels.

src/lsp/cobol_ptree/terms.ml
```diff
@@ -269,10 +269,42 @@ and class_ =
...
+  | CallNumvalC of ident_or_nonnum list
+  | CallLocaleDate of locale_datetime_args
+  | CallLocaleTime of locale_datetime_args
+  | CallLocaleTimeFromSeconds of locale_datetime_args
+  | CallFormattedDatetime of formatted_datetime_args
+  | CallFormattedTime of formatted_datetime_args
+
+and call_trim_args = {
+  trimmed: effective_arg;
+  position: leading_trailing option;
```
nberth, line 287: The name of the `leading_trailing` type is confusing. `tip: trimming_tip` could me more appropriate.

src/lsp/cobol_ptree/terms.ml
```diff
@@ -269,10 +269,42 @@ and class_ =
...
+  | CallLength of call_length_args
+  | CallNumvalC of ident_or_nonnum list
+  | CallLocaleDate of locale_datetime_args
+  | CallLocaleTime of locale_datetime_args
+  | CallLocaleTimeFromSeconds of locale_datetime_args
+  | CallFormattedDatetime of formatted_datetime_args
+  | CallFormattedTime of formatted_datetime_args
+
+and call_trim_args = {
+  trimmed: effective_arg;
```
nberth, line 286: `arg` in the inline record.

src/lsp/cobol_ptree/terms_visitor.ml
```diff
@@ -232,9 +250,42 @@ and fold_address (v: _ #folder) =
 
 and fold_inline_call (v: _ #folder) =
   handle v#fold_inline_call
-    ~continue:begin fun { call_fun; call_args } x -> x
-      >> fold_name' v call_fun
-      >> fold_list ~fold:fold_effective_arg v call_args
+    ~continue:begin function
```
nberth, line 253: In this case you should `begin fun c x -> match c with` and then `<pattern> -> x` for clarity.

src/lsp/cobol_ptree/terms_visitor.ml
```diff
@@ -286,6 +337,10 @@ and fold_expr (v: _ #folder) =
 and fold_expr' (v: _ #folder) =
   handle' v#fold_expr' ~fold:fold_expr v
 
+and fold_leading_trailing (_: _ #folder) : leading_trailing -> 'a -> 'a = function
```
nberth, line 340: Here you don't use the `fold_leading_trailing` method: `= leaf v#fold_leading_trailing` should be sufficient.

src/lsp/cobol_typeck/old_env_builder.ml
```diff
@@ -98,7 +98,8 @@ let for_compilation_unit =
       let name = Cobol_ptree.program_name' name in
       acc_result @@ fun parent_env -> build_env name env ?parent_env
     method! fold_function_unit f =
-      acc_result @@ fun _ -> build_env f.function_name f.function_env
+      acc_result @@ fun _ -> build_env
```
nberth, line 101: I'd suggest to take inspiration from `Cobol_ptree.program_name'` and maybe have a `Cobol_ptree.function_name'`.

test/output-tests/syn_functions.expected
```diff
@@ -81,12 +81,22 @@ syn_functions.at-198-prog.cob:16.11-16.15:
...
+syn_functions.at-198-prog.cob:16.24:
   13          01  ret PIC X.
   14              
   15          PROCEDURE DIVISION RETURNING ret.
   16 >            MOVE FUNCTION x TO ret
-----                              ^^
+----                           ^
+  17              MOVE FUNCTION x TO ret
+  18              .
+>> Hint: Missing <word>
```
nberth, line 92: There's something fishy here.

src/lsp/cobol_parser/parser_engine.ml
```diff
@@ -284,6 +285,20 @@ let post_production ({ preproc = { tokzr; _ }; _ } as ps)
...
+    | IntrinsicFunctionSpecifier intrinsics ->
+        let intrinsics = List.map (~&) intrinsics in
+        let tokzr, token, tokens
+          = Tokzr.intrinsic_functions_specifier ~intrinsics tokzr token tokens in
+        update_tokzr ps tokzr, token, tokens
+    | IntrinsicFunctionAllSpecifier ->
+        let tokzr, token, tokens
+          = Tokzr.intrinsic_functions_specifier tokzr token tokens in
+        update_tokzr ps tokzr, token, tokens
+    | _ ->
```
nberth, line 299: Please list the missing cases here. Also, extract this to a smaller function above `register_intrisics ?intrinsics tokzr token tokens = …`, and avoid the nested pattern:
```ocaml
| Post_function_specifier IntrinsicFunctionSpecifier intrinsics ->
    …
| Post_function_specifier IntrinsicFunctionAllSpecifier ->
    …
```

src/lsp/cobol_parser/text_lexer.mli
```diff
@@ -34,12 +34,14 @@ val punct_of_token : (Grammar_tokens.token, string) Hashtbl.t
...
 (* --- *)
 
-val create: ?decimal_point_is_comma:bool -> unit -> lexer
+val create: ?decimal_point_is_comma:bool -> EzCompat.StringSet.t -> lexer
 val handle_of_token: lexer -> Grammar_tokens.token -> token_handle
 val reserve_words: lexer -> Cobol_config.words_spec -> unit
 val enable_tokens: TokenHandles.t -> unit
 val disable_tokens: TokenHandles.t -> unit
 val decimal_point_is_comma: lexer -> lexer
+val intrinsic_functions_specifier: ?intrinsics:string list -> lexer -> lexer
```
nberth, line 43: `val register_intrinsics: …`

src/lsp/cobol_config/cobol_config.ml
```diff
@@ -21,6 +21,7 @@ include Types
 module Options = Options
 module Default = Default
 module Diagnostics = Config_diagnostics
+module Reserved = Reserved_words
```
nberth, line 24: Do we need to export that?

src/lsp/cobol_parser/text_lexer.ml
```diff
@@ -99,17 +102,42 @@ let silenced_keywords =
 
 (* --- *)
 
-let create ?(decimal_point_is_comma = false) () =
+(*TODO: Add from config + cmd line*)
+let create ?(decimal_point_is_comma = false) intrinsics =
```
nberth, line 106: Please keep using a label for `instrisics`, as the type is not explicit enough.

src/lsp/cobol_parser/text_lexer.ml
```diff
@@ -99,17 +102,42 @@ let silenced_keywords =
...
   {
     lexer with decimal_point_is_comma = true;
   }
 
+let intrinsic_functions_specifier ?intrinsics lexer =
+  match intrinsics with
+  | Some intrinsics ->
+    {
+      lexer with intrinsic_functions_specifier =
+        List.fold_left (fun acc intrinsic ->
```
nberth, line 124: Isn't this a `StringSet.inter` or something similar?

src/lsp/cobol_parser/text_tokenizer.ml
```diff
@@ -54,7 +54,7 @@ let combined_tokens =
     DATA_RECORDS, "DATA_RECORDS";
     NEXT_PAGE, "NEXT_PAGE";
     NEXT_SENTENCE, "NEXT_SENTENCE";
-  ]
+  ] @ List.map (fun (fst, snd) -> snd, fst^"_FUNC") Text_keywords.intrinsic_functions)
```
nberth, line 57: ??? Why would be add *ad hoc* `_FUN` words in that table of combined tokens?

src/lsp/cobol_parser/text_tokenizer.ml
```diff
@@ -153,6 +154,183 @@ let preproc_n_combine_tokens ~source_format =
   let rec skip ((p', l', dgs) as acc) ((p, l) as pl) = function
     | 0 -> acc, pl
     | i -> skip (hd p :: p', hd l :: l', dgs) (tl p, tl l) (i - 1)
+  and subst_function ?(in_repo = false) ((p', l', dgs) as acc) ((p, l) as pl) =
```
nberth, line 157: Surely some more help from the lexer will help keep that clean.

src/lsp/cobol_parser/text_tokenizer.ml
```diff
@@ -244,13 +467,15 @@ let preproc_n_combine_tokens ~source_format =
     | [CONSTANT]                     -> Error `MissingInputs
     | CONSTANT :: RECORD :: _          -> subst_n CONSTANT_RECORD 2
 
-    | [PROGRAM_ID]
-    | [PROGRAM_ID; PERIOD]           -> Error `MissingInputs
-    | PROGRAM_ID :: PERIOD :: _        -> info_word_after 2
-    | PROGRAM_ID :: _                  -> info_word_after 1
+    | [(PROGRAM_ID | FUNCTION_ID)]
```
nberth, line 470: For readability of each case, please duplicate entries instead of using parentheses here.

src/lsp/cobol_parser/text_tokenizer.ml
```diff
@@ -529,6 +759,8 @@ type lexer_update =
   | Enabled of Text_lexer.TokenHandles.t
   | Disabled of Text_lexer.TokenHandles.t
   | CommaBecomesDecimalPoint
+  | IntrinsicFunctionsSpecifier of string list option
```
nberth, line 762: `RegisterIntrincsics`

nberth (commented):
Some quick comments for now. I think the part in the text lexer/tokenizer can be improved.

nberth:
Further thoughts. I think this PR goes into the right direction, but some points need to be discussed and addressed to achieve a more manageable solution:
- All considering, I don't see any good reason to combine `FUNCTION` + `<intrinsic_name>` tokens. The `FUNCTION` can just be made optional in front of intrinsic function tokens in the grammar. The combined tokens are harder to deal with during re-tokenization, so it's better to avoid them whenever possible;
- I suggest to use a new hashtable of intrinsic function tokens in the `Text_lexer`, similar to the one used for keywords (with a mutable flag that indicates the intrinsic is available without `FUNCTION`) — even, reusing the same table should be doable, by tweaking the type of its entries with a record `{as_keyword; optional_token; as_intrinsic: intrinsic_handle}`. The mechanics for registering those should be inversible (for parser rewinds), and for efficiency intrinsic tokens should be manipulated via handles like keyword tokens;
- Please disregard my comment about "specifier"; I did see it used in the standard since then.

src/lsp/cobol_parser/grammar_tokens.mly
emilienlemaire: But for the case of a `CallFunc` we need a `name with_loc`, I can change it to `UserDefined of name with_loc | Intrinsic of intrinsic_function_identifier`

src/lsp/cobol_parser/grammar.mly
emilienlemaire: Same as for the typing of the token, we would need a special case in the type of `CallFunc {call_fun: <special_type_here>; ...}` if that's what you'd like

src/lsp/cobol_parser/grammar_tokens.mly
nberth: What is this token used for actually? I would have thought it's something like a `used-defined-intrinsic-name` or something like that. If it's user-defined, the scanner/tokenizer should have the mapping to the actual intrinsic token already, and therefore this one wouldn't be useful.

emilienlemaire: It's used in `function_identifier` rule of the grammar to allow not having `FUNCTION` in front of the intrinsic name if the name was specified in the REPOSITORY. Wither we have this token, for intrinsic function without special handling of arguments.
If we don't have this token, we should have one token by intrinsic function name, which just adds more matching in the grammar to handle this special case

nberth: I see. So this should be something like `GENERIC_INTRINSIC`, not carrying a name, but with a piece of AST that references the intrinsic. These identifiers should all be known in advance, so we can have a dedicated ADT for them.

src/lsp/cobol_parser/grammar.mly
nberth: Yes. But not `CallFunc`, rather `CallGenericIntrinsic of { intrinsic: generic_intrinsic_identifier; args: arguments }`.

src/lsp/cobol_ptree/terms.ml
```diff
@@ -694,9 +892,64 @@ module FMT = struct
...
     | ProgAddress i -> fmt "ADDRESS@ OF@ PROGRAM@ %a" ppf pp_term i
 
-  and pp_inline_call ppf { call_fun; call_args } =
-    fmt "FUNCTION@ %a@ @[<1>(%a)@]" ppf pp_name' call_fun
-      (list ~sep:comma pp_effective_arg) call_args
+  and pp_inline_call ppf f =
+    match f with
+    | CallFunc { call_fun; call_args } ->
+      fmt "FUNCTION@ %a@ @[<1>(%a)@]" ppf pp_name' call_fun
+        (list ~sep:comma pp_effective_arg) call_args
```
nberth, line 899: Just noticed this: I think `~sep:comma` should be `~sep:sp` instead (commas are like space separators).

src/lsp/cobol_parser/grammar.mly
```diff
@@ -2253,9 +2260,38 @@ let qualident ==
...
 
+let arg_par(X) ==
+  "("; ~=X; ")"; < >
+
 let function_ident ==
- | FUNCTION; n = function_name; al = arguments; { { call_fun = n; call_args = al } }
- | FUNCTION; n = function_name;                 { { call_fun = n; call_args = [] } }
+ | FUNCTION; n = function_name; al = arguments; { CallFunc { call_fun = n; call_args = al } }
+ | FUNCTION; n = function_name;                 { CallFunc { call_fun = n; call_args = [] } }
+ | FUNCTION?; n = loc(intrinsic_function_name); al = arguments;
```
nberth, line 2269: Have you tried with a rule `FUNCTION?; ~ intrinsic_call; < >`?

src/lsp/cobol_parser/grammar_tokens.mly
```diff
@@ -1399,4 +1399,127 @@
 %token XML_NTEXT                    [@keyword.silenced]
 %token XML_TEXT                     [@keyword.silenced]
 
+(* INTRINSIC FUNCTIONS *)
+%token <Cobol_ptree.generic_intrinsic_identifier> INTRINSIC_FUNC
+
+(* Intrinsic functions that need special care (either because their name is a COBOL keyword or
+   because they have special arguments) *)
+%token BYTE_LENGTH_FUNC                  [@intrinsic.unique "BYTE-LENGTH"]
```
nberth, line 1407: You could further simplify these declarations by using the same trick as for keywords (that uses the name of the token and replaces `_` with `-` when the `@keyword` has no payload).

nberth: (At least for the non-unique)

src/lsp/cobol_parser/text_lexer.ml
```diff
@@ -24,12 +24,24 @@ module TYPES = struct
       mutable reserved: bool;
       mutable enabled: bool;
     }
-  and token_handle = optional_token
+  and keyword_handle = optional_token
 
+  type intrinsic_handle =
+    {
+      token: Grammar_tokens.token;
+      mutable specified: bool;
```
nberth, line 32: What do you mean by `specified`? Is it "registered", without need to `FUNCTION` prefix?

src/lsp/cobol_parser/text_lexer.ml
```diff
@@ -24,12 +24,24 @@ module TYPES = struct
...
+    }
+
+  type token_handle = {
+    as_keyword: keyword_handle option;
+    as_intrinsic: intrinsic_handle option;
+  }
   type lexer =
     {
-      token_of_keyword: (string, token_handle) Hashtbl.t;
+      token_of_keyword: (string, keyword_handle) Hashtbl.t;
```
nberth, line 41: Why do we now have three tables instead of just one before?

src/lsp/cobol_parser/text_lexer.ml
```diff
@@ -39,26 +51,58 @@ module TokenHandles = struct
   include Set.Make
```
nberth, line 51: Not sure `TokenHandles` applies to intrinsics.

src/lsp/cobol_parser/text_lexer.ml
```diff
@@ -24,12 +24,24 @@ module TYPES = struct
...
-  and token_handle = optional_token
+  and keyword_handle = optional_token
 
+  type intrinsic_handle =
+    {
+      token: Grammar_tokens.token;
+      mutable specified: bool;
+    }
+
+  type token_handle = {
```
nberth, line 35: Having second thoughts about that record. Maybe it's a matter of changing `enabled: bool` into `enabled: No | As_keyword | As_registered_intrinsic | Both(?)`?

src/lsp/cobol_parser/text_lexer.mli
```diff
@@ -40,6 +46,7 @@ val reserve_words: lexer -> Cobol_config.words_spec -> unit
 val enable_tokens: TokenHandles.t -> unit
 val disable_tokens: TokenHandles.t -> unit
 val decimal_point_is_comma: lexer -> lexer
+val specify_intrinsic: lexer -> string -> unit
```
nberth, line 49: `val register_as_intrinsics: TokenHandles.t -> unit` (maybe `IntrinsicHandles.t` instead, if you want some more typing constraints over tokens). Missing `unregister_intrinsics: … -> unit` as well.
Or something along those lines. Possibly the only changes needed to this `mli`: `as_intrinsic` and `token_of_intrinsic_handle` should not be needed.

src/lsp/cobol_parser/text_tokenizer.ml
nberth: This part should be removed or much simplified: we need to avoid combining `FUNCTION` tokens if at all possible.

src/lsp/cobol_parser/text_tokenizer.ml
```diff
@@ -187,6 +315,51 @@ let preproc_n_combine_tokens ~source_format =
         consume_comment ~loc:(hd l) ~revtoks:[] ~at_end
           Comment_entry acc suff
     in
+    let function_specifier = function
```
nberth, line 318: Same. Maybe the changes need to be handled in the same way as picture strings in the other function below.

test/output-tests/syn_functions.expected
emilienlemaire: That might be because we are in a function definition, when I change `FUNCTION-ID` and `END FUNCTION` by `PROGRAM-ID` and `END PROGRAM` in the source file, the syntax errors are no more present

nberth: Yes we can ignore those errors: the current grammar only accepts more programs after having parsed a first program (*i.e* not a function definition, a class, etc).

nberth:
@emilienlemaire I think I have a state with clean handling of rewinds. I'll split that into separate commits and then I think this PR will be ready.

nberth (approved):

PR #263, opened 2024-04-08
Properly resolve the default debugger launch configuration

### LINKED ISSUE #219: Implicit build task should be the debug build task
So that launching debugging when there is no `tasks.json` file works.

I.e the debug build task in `tasks.json` should contain:
```
			"group": {
				"kind": "build",
				"isDefault": true
			}
```

That should probably be handled in `attributes_spec` (`superbol_tasks.ml`).

nberth:
Fixes #219


ddeclerck (approved):
Great !

PR #265, opened 2024-04-10
Use per-language scanners for `EXEC`/`END-EXEC` blocks

nberth:
Also add the ability to pass some state upon each successive scan of blocks for the same language.


emilienlemaire (approved):
All good for me :+1:

ddeclerck (approved):
Tested and approved ;)

PR #266, opened 2024-04-11
Bugfix: handle contexts that terminate upon shift instead of reduce

nberth:
Fixing that was an epic fight :sweat_smile:.

Before these changes, some contexts were not exited properly. For instance `01 x CONSTANT 12345.` entered context `constant`, but did not leave it. The reason was simple: this context is here left via a shift instead of a reduction, which was previously though as the only means for a grammar rule to be left. The latter assumption turned out to be wrong, and this PR addresses this issue.

Prior to this PR, the new test failed, with the `BYTE-LENGTH` words wrongly recognized as keywords.

### DIFF
--- a/src/lsp/cobol_parser/context/gen_context.ml
+++ b/src/lsp/cobol_parser/context/gen_context.ml
@@ -71,8 +71,16 @@ let emit_nonterminal_contexts ppf =
   Fmt.pf ppf "\
     \  | _ -> None\n"
 
+let pp_context ppf s =
+  Fmt.string ppf (String.capitalize_ascii s)
+
 let pp_contexts =
-  Fmt.(list ~sep:(any ";@ ") (fun ppf s -> pf ppf "%s" (String.capitalize_ascii s)))
+  Fmt.(list ~sep:(any ";@ ") pp_context)
+
+let context_of_production =
+  Production.tabulate begin fun prod ->
+    Option.map fst @@ nonterminal_context (Production.lhs prod)
+  end
 
 let emit_contexts_mapping ppf =
   Fmt.pf ppf "\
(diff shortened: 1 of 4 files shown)


ddeclerck:
> Fixing that was an epic fight 😅.

Indeed it seems you have struggled ;)

> Before these changes, some contexts were not exited properly. For instance `01 x CONSTANT 12345.` entered context `constant`, but did not leave it. The reason was simple: this context is here left via a shift instead of a reduction, which was previously though as the only means for a grammar rule to be left. The latter assumption turned out to be wrong, and this PR addresses this issue.

Just out of curiosity: are there many instances of such contextes left via a shift ? Why does it even happen ?

ddeclerck (approved):

nberth:
> Just out of curiosity: are there many instances of such contextes left via a shift ? Why does it even happen ?

For our grammar, the amount of *potential* context-leaving shifts I've found is not that small. However, those may be rare in practice for actual programs.

These transitions appear when, from a given state in the grammar, a production `p` in a context (grammar rule) is a strict prefix of another production `p'` that is not in the context (it is defined in another rule). In the case of `CONSTANT` above, we have something like (much simplified):
```ocaml
let constant_spec :=
  | CONSTANT; expression;
  | constant_value_length; name;

let constant_value_length [@context constant] :=
  | CONSTANT; BYTE_LENGTH;
```
The context is actually entered when "entering" `constant_spec` (not `constant_value_length`) when reading `CONSTANT`,  but needs to be existed upon a shift if it then reads anything else than `BYTE-LENGHT`.

ddeclerck:
I see. That's tricky indeed.

PR #267, opened 2024-04-17
Worldcities fixes

emilienlemaire:
This should enable the full parsing of world cities (except for `>>D` compiler directives).
For the `special-names` error, we must move the paragraph before the `repository` paragraph.


nberth:
Thanks for those changes. I first need to deal with #262 prior to that one.

emilienlemaire:
This PR is for @lefessan, to parse world cities for the presentations to come

lefessan:
@nberth I would like to have these PRs on worldcities merged asap so that I can work on a demo for the meetup.

nberth:
> @nberth I would like to have these PRs on worldcities merged asap so that I can work on a demo for the meetup.

I'll be working on this as soon as I can. I need to fix a few things.

emilienlemaire:
No, I made this one out of commits from #262 that were compiling and working fine (but not very good codewise) so we could use this branch for the presentation, I think it can be closed now that #262 is merged.
The only other problems concerning world cities are the `>>D` compiler directive and the order of the paragraphs in the environment division (worldcities uses `repository` before `special-names` which is not allowed by our parser).

PR #279, opened 2024-05-03
Better handling of `,` & `;` separators in manipulated text

nberth:
These changes shall enable `EXEC`/`END-EXEC` blocks with languages in which commas and semicolons are relevant (like SQL).  This is achieved by adding a new `Separator` constructor for elements of manipulated text.

Further handling of weird GC extensions that allow `77,x;PIC,99,99.` (equivalent to `77 x PIC 99,99.`) required some rewriting of the scanning code.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,5 @@
-dune             linguist-generated
 linking_flags.sh linguist-generated
+/dune            linguist-generated
 /.drom           linguist-generated
 /Makefile        linguist-generated
 /dune-project    linguist-generated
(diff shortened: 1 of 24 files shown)


.gitattributes
```diff
@@ -1,4 +1,3 @@
-dune             linguist-generated
```
emilienlemaire, line 1: Isn't this file still generated?

emilienlemaire (approved):
All good, just a small question that can be fixed easily

.gitattributes
nberth: Ah you meant `/dune` (didn't notice there was one). I'll fix that.

PR #280, opened 2024-05-03
Add `ID` keyword, synonym for `IDENTIFICATION` (GC extension)

nberth:


test/output-tests/run_ml.expected
```diff
@@ -1,4 +1,14 @@
...
+run_ml.at-122-prog.cob:21.13:
+  18          01  0SpecialTAGName  PIC X(3) VALUE "abc".
+  19   
+  20          01  employee.
+  21 >            05 id            PIC 9(1)  value 1.
+----                ^
+  22              05 name          PIC X(10) value "Someone".
+  23              05 dept          PIC X(10) value "Marketing".
+>> Hint: Missing .
+
```
emilienlemaire, line 11: Are those errors expected?

test/output-tests/run_ml.expected
```diff
@@ -9,925 +19,99 @@ run_ml.at-122-prog.cob:21.14-21.16:
...
+>> Hint: Missing DIVISION
 
-run_ml.at-122-prog.cob:89.42-89.46:
+run_ml.at-122-prog.cob:89.53-89.55:
   86              .
   87   
   88              *> Another test with mixed attributes and values
   89 >            XML GENERATE out FROM EMPLOYEE TYPE OF ID IS ATTRIBUTE
-----                                             ^^^^
+----                                                        ^^
```
emilienlemaire, line 97: This also seems like an unwanted new error

emilienlemaire (commented):
It seems like a few new errors got in, if you deem it ok to ignore them I'll approve the PR

test/output-tests/run_ml.expected
nberth: Yes, the actual command in the autotest file passes `-fnot-reserved=id`. We currently don't parse these arguments.

test/output-tests/run_ml.expected
nberth: Same reason as above.

emilienlemaire (approved):

PR #281, opened 2024-05-03
Enable coverage file generation

emilienlemaire:


README.md
```diff
@@ -198,6 +198,15 @@ Press <kbd>F10</kbd> to step to the next statement, or <kbd>F5</kbd>
 again to continue until the next breakpoint, or termination of the
 program.
 
+## Coverage
+
+To enable coverage you can set `enable-coverage` to `true` in your
```
nberth, line 203: Could you reformulate into two distinct paragraphs that focus on one element each?
The first one focuses on coverage files, the second one on highlighting in the interface.

In addition, for now the `README.md` does not mention `task.json` or `launch.json` (I think), as most features described should work by default without those files (this is for ease of access to VS Code beginners). If we don't enable coverage by default (maybe that's a good idea to have to enable it), then we need paragraphs stating that those files need to be constructed via (`Run > Configure build tasks…` or similar commands).

src/vscode/superbol-vscode-platform/superbol_tasks.ml
```diff
@@ -111,9 +113,9 @@ let cobc_execution ?config attributes =
       ~mk:((^) "-fformat=") |>
     attr_bool_flag "for-debug" ~attributes
       ~ok:(fun args -> "-fsource-location" :: "-ftraceall" ::
-                       "-g" ::
-                       "-Q" :: "--coverage" ::
-                       "-A" :: "--coverage" :: args) |>
+                       "-g" :: args) |>
+    attr_bool_flag "enable-coverage" ~attributes
+      ~ok:(fun args -> "--coverage" :: args) |>
```
nberth, line 118: Don't we need `-A --coverage` and `-Q --coverage` still (not sure…)?
```
$ cobc --help
…
  -A <options>          add <options> to the C compile phase
  -Q <options>          add <options> to the C link phase
  --coverage            instrument generated binaries for coverage
…
```

nberth (changes requested):
I think the `README.md` now needs to introduce some of the VS Code config files.

src/vscode/superbol-vscode-platform/superbol_tasks.ml
emilienlemaire: No they are equivalent, maybe these `-Q --coverage -A --coverage` comes from times where `--coverage` was not a direct GnuCOBOL options, and it had to be passed to the C compiler and linker by hand

README.md
emilienlemaire: https://github.com/OCamlPro/superbol-studio-oss/blob/e2bd9a73ec3c94dd4792d3867e441e63077ffc68/README.md?plain=1#L173
The `README.md` does mention the `task.json` file, I'll edit the paragraph to focus on both features in distinct way tho

nberth: `Setting up a Build Task for Debugging` is out of date: by default we don't need the `task.json` anymore.

nberth: You can probably reuse part of that paragraph down in `Coverage`.

emilienlemaire: So I should delete `Setting up a Build Task` paragraph? I think maybe, if the README is outdated, a full PR to update it should be the way to go.
It would be better than modifying parts of it when we notice the deprecation

nberth:
No I think it is "globally" up-to-date.  Here it's just lifting of a limitation (mandatory presence of `task.json` for debug) that was not reflected in there.

nberth (commented):
Minor reformulations

README.md
```diff
@@ -198,6 +198,32 @@ Press <kbd>F10</kbd> to step to the next statement, or <kbd>F5</kbd>
...
+
+### Display coverage hightlighting
+
+After running a program and having its coverage files generated
+you might want to display its content on a line by line basis
+in your program source code.
+
+To enable this coverage highlighting you can set `coverage` to
+`true` in your `launch.json` file. If you do not have this
+file yet, you can generate it by selection the `Run and Debug`
```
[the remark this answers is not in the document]
emilienlemaire: I don't understand what you mean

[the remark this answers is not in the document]
nberth: Found it. `Run > Add Configuration…` automatically creates the `launch.json` if it does not exist; `Terminal > Configure Tasks…` for `tasks.json`.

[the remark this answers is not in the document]
nberth: (noticed `coverage` seems `true` by default in `"SuperBOL: debug (launch)"` configuration.

README.md
```diff
@@ -198,6 +198,32 @@ Press <kbd>F10</kbd> to step to the next statement, or <kbd>F5</kbd>
...
+`task.json` file.
+
+The coverage files generated are in `gcov` format, so they are
+portable and you can use them as you would any other coverage
+files from other languages.
+
+### Coverage Hightlighting
+
+After running a program and having its coverage files generated.
+You might want to display their content on a line by line basis
```
nberth, line 216:
```suggestion
You might want to display their contents on a line by line basis
```

nberth (approved):

PR #288, opened 2024-05-21
Bump menhir version + update merlin submodule

emilienlemaire:
This has been tested with menhir versions `20230415` and `20231231` and works fine


nberth (approved):
Great! Thanks!

PR #292, opened 2024-05-24
Support diagnostics returned by scanners for `EXEC`/`END-EXEC` blocks

nberth:


emilienlemaire (approved):
All good for me :100:

PR #299, opened 2024-06-03
cobol_indent : scanner improvements, to be used for cartography

lefessan:
Some improvements to return more information in case this scanner is used instead of the COBOL parser to retrieve faster information (but less exact)

### DIFF
--- a/src/lsp/cobol_indent/config.ml
+++ b/src/lsp/cobol_indent/config.ml
@@ -53,6 +53,9 @@ let cobolx_format = {
 }
 
 let default = {
+  scan_for_indent = true;
+  verbosity = 1;
+
   source_format = fixed_format ;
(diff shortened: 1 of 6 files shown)


import/merlin
nberth, line 1: You probably need to `git submodule update`.

lefessan: Chez moi, ça passe, du coup, je ne comprends pas vraiment pourquoi ça échoue sur le CI...

nberth: Commit b3de92cd4fb23 reverts commits from submodules. You just need to edit that commit and restore the changes to submodules.

lefessan: What I don't understand is why it works locally on my computer and not on the CI, it should fail on my computer too, no ?

nberth (approved):

PR #302, opened 2024-06-05
Add rudimentary devices to enable inspection of the parser state

nberth:
Mostly adds internal support for inspecting the parser's state right before EOF; adapts and fixes the source overlay manager along the way.


emilienlemaire (approved):

src/lsp/cobol_common/srcloc.ml
```diff
@@ -273,6 +273,10 @@ let rec in_area_a: srcloc -> bool = function
   | Rpl { in_area_a; _ } -> in_area_a
   | Cat { left; _ } -> in_area_a left
 
+let is_pointwise: srcloc -> bool = function
+  | Raw (s, e, _) -> s.pos_cnum == e.pos_cnum && s.pos_fname = e.pos_fname
```
ddeclerck, line 277: Why physical equality ?

ddeclerck (approved):

src/lsp/cobol_common/srcloc.ml
nberth: Well. `cnum`s are `int`s, so it does not really matter.

PR #304, opened 2024-06-10
Context-sensitive completion for keywords and user-defined words

NeoKaios:
Added back completion in LSP to provide context specific completion items
Allow completion of keywords thanks to grammarware gen_expect.ml and completion for variable name (qualified name or not) for procedure and zones thanks to qualmap


src/lsp/cobol_parser/grammar_tokens.mly
```diff
@@ -1018,37 +1018,37 @@
...
+%token NEXT_SENTENCE                [@keyword.combined]
 
 (* Combined DATA tokens *)
-%token DATA_RECORD
-%token DATA_RECORDS
+%token DATA_RECORD                  [@keyword.combined]
+%token DATA_RECORDS                 [@keyword.combined]
 
 (* Combined CONSTANT *)
 %token CONSTANT_RECORD
```
nberth, line 1054: Did you leave that one without `[@keyword.combined]` on purpose?

src/lsp/cobol_parser/keywords/gen_keywords.ml
```diff
@@ -78,10 +82,13 @@ let pp_terminal ppf t =
 let cobolize name =
   String.map (function '_' -> '-' | c -> c) name
 
+let combined_cobolize name =
```
nberth, line 85: `combined` will suffice.

src/lsp/cobol_parser/grammar_tokens.mly
NeoKaios: Yes, it wasn't in the original keyword combined list
https://github.com/OCamlPro/superbol-studio-oss/pull/304/files#diff-788d7917aebbdc3ba4afb53fef4f703a4feafe84fa888138c71355ed1d4d98ce

src/lsp/cobol_parser/keywords/gen_keywords.ml
NeoKaios:
```ocaml
let combined attrs =
  List.find_opt (Attribute.has_label "keyword.combined") attrs |>
  Option.map Attribute.payload
```
There is already a function with that name :upside_down_face:

test/lsp/lsp_completion.ml
```diff
@@ -0,0 +1,631 @@
...
+      position.line position.character;
+    match LSP.Request.complete server params with
+    | None ->
+        Pretty.out "Failed completion@."
+    | Some `CompletionList { items; _ } when items == [] ->
+        Pretty.out "Empty completion list@."
+    | Some `CompletionList { items; _ } ->
+        let pp_comp_item ppf (item: CompletionItem.t) =
+          Fmt.string ppf item.label in
+        Pretty.out "List of completions (%d entries): [%a]\n" (List.length items) (Fmt.list ~sep:(Fmt.any ";") pp_comp_item) items;
```
nberth, line 41:
```suggestion
        Pretty.out "List of completions (%d entries): [%a]\n"
          (List.length items) (Fmt.list ~sep:(Fmt.any ";") pp_comp_item) items;
```

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -0,0 +1,421 @@
...
+(* This source code is licensed under the GNU Affero General Public       *)
+(* License version 3 found in the LICENSE.md file in the root directory   *)
+(* of this source tree.                                                   *)
+(*                                                                        *)
+(**************************************************************************)
+
+module NEL = Cobol_common.Basics.NEL
+
+let cmlyname = ref None
+let external_tokens = ref ""
```
nberth, line 17: Do you need that in this script?

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -0,0 +1,421 @@
...
+let nullable_nonterminals = Lr1.tabulate begin fun lr1 ->
+  SymbolSet.from_list_map ~add:SymbolSet.add_option
+  (fun (prod, idx) ->
+    match (Production.rhs prod).(idx) with
+      | N nt, _, _ when Nonterminal.nullable nt -> Some (N nt)
+      | _ | exception Invalid_argument _ -> None)
+    (Lr0.items @@ Lr1.lr0 lr1)
+  end
+
+let reducable_productions = Lr1.tabulate begin fun lr1 ->
```
nberth, line 193:
```suggestion
let reducible_productions: lr1 -> production list =
  Lr1.tabulate begin fun lr1 ->
```

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -0,0 +1,421 @@
...
+    | (prod, idx) when Array.length (Production.rhs prod) == idx ->
+        Some prod
+    | _ -> None)
+  (Lr0.items @@ Lr1.lr0 lr1)
+  |> List.sort_uniq Production.compare
+end
+
+let emit_accaptable_nullable_nonterminals_in_env ppf =
+  Fmt.pf ppf {|
+let accaptable_nullable_nonterminals_in ~env : Menhir.xsymbol list =
```
nberth, line 204:
```suggestion
let acceptable_nullable_nonterminals_in ~env : Menhir.xsymbol list =
```

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -0,0 +1,421 @@
...
+  CompEntryMap.empty
+  |> inv (CompEntryMap.fold (fun s l acc -> (l, CompEntrySet.elements s)::acc)) []
+  |> pp_match_cases ppf
+    Fmt.(using Lr1.to_int int)
+    (pp_brackets @@ pp_list pp_completion_entry)
+    "[])"
+
+let guess_default_value ~mangled_name ~typ =
+  match typ, mangled_name with
+  | _ when EzString.ends_with ~suffix:"option" typ ||
```
nberth, line 284:
```suggestion
  | _ when EzString.ends_with ~suffix:" option" typ ||
```
(Same for the other suffices below)

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -0,0 +1,421 @@
...
+          EzString.starts_with ~prefix:"option_" mangled_name
+  -> Some "None"
+  | _ when EzString.ends_with ~suffix:"list" typ ||
+  EzString.starts_with ~prefix:"loption_" mangled_name ||
+  EzString.starts_with ~prefix:"list_" mangled_name
+  -> Some "[]"
+  | _ when EzString.ends_with ~suffix:"unit" typ
+  -> Some "()"
+  | _ when EzString.ends_with ~suffix:"bool" typ ||
```
nberth, line 293: It's probably an equality test that we want for `bool` and `unit`.

Why the `match` if you never desctruct a value (*i.e,* every pattern is `_`)?

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -0,0 +1,421 @@
...
+  let given_default = Nonterminal.attributes nt
+    |> List.find_opt (Attribute.has_label "default")
+    |> Option.map Attribute.payload in
+  match given_default with
+    | Some _ -> given_default
+    | None -> guessed_default
+  end
+
+let emit_default_value_of_nullables ppf =
+  Fmt.pf ppf {|
```
nberth, line 322: You can likely use `Fmt.string` directly here.

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -0,0 +1,421 @@
...
+end
+
+
+let () =
+  let ppf = Fmt.stdout in
+  Fmt.pf ppf
+    "(* Caution: this file was automatically generated from %s; do not edit *)\
+     \nmodule NEL = Cobol_common.Basics.NEL\
+     \nmodule Menhir = Grammar.MenhirInterpreter\
+     \nopen Grammar_tokens@\n\n"
```
nberth, line 408: That might be where `!external_tokens` is needed. You could also require a mandatory `--nel-module` argument for `NEL` (in case that script is re-used somewhere else, for another parser).

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -0,0 +1,421 @@
...
+  match given_default with
+    | Some _ -> given_default
+    | None -> guessed_default
+  end
+
+let emit_default_value_of_nullables ppf =
+  Fmt.pf ppf {|
+let guessed_default_value_of_nullables (type a): a Menhir.symbol -> a = function
+  (* If this function does not compile it probably means you're missing a
+     [@default] attribute on a nullable nonterminal in the grammar *)
```
nberth, line 325: Does this comment still hold (there'e a wildcard with `raise Not_found`) ?

nberth: I'm also wondering why this function takes a `symbol`, whereas it only makes sense on `nonterminal`.

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -0,0 +1,421 @@
...
+  List.find_opt (Attribute.has_label "completion") |>
+  Option.map Attribute.payload |>
+  inv Option.bind (fun s -> Some (Custom s))
+
+let pp_xsymbol_of_nt ppf nonterminal =
+  Fmt.pf ppf "X(N N_%s)" (Nonterminal.mangled_name nonterminal)
+
+(* --- *)
+
+module SuperSet (OrderedType : Set.OrderedType) = struct
```
nberth, line 94: Why "Super"? It's just a regular `Make_set`.

- Ok for `add_option`;
- `add_list` should be removed and `Set.union s (Set.of_list l)` should be used instead;
- `from_list_map` is a bit far-fetched with its `add` labeled argument…

src/lsp/cobol_parser/expect/gen_expect.ml
nberth, line 1: That file deserves a pass of `ocp-indent`.

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -0,0 +1,421 @@
...
+    let compare entry1 entry2 =
+      match entry1, entry2 with
+        | Custom s1, Custom s2 -> String.compare s2 s1
+        | K nel1, K nel2 -> NEL.compare Terminal.compare nel1 nel2
+        | Custom _, K _ -> -1
+        | K _, Custom _ -> 1
+  end)
+end
+
+module Map5 (OrderedType : Map.OrderedType) = struct
```
nberth, line 125: Why `5`? Also this is a `Make_map`

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -0,0 +1,421 @@
...
+module SymbolMap = Map5(struct
+  type t = SymbolSet.t
+  let compare = SymbolSet.compare
+end)
+
+(* --- *)
+
+let emit_pp_completion_entry ppf custom_types = (* For debug *)
+  Fmt.pf ppf "\nlet pp_completion_entry: completion_entry Fmt.t = fun ppf -> function\n";
+  Fmt.pf ppf "  | K tokens -> Fmt.pf ppf \"%%a\" (Fmt.list ~sep:Fmt.sp Fmt.string) (List.map Grammar_printer.print_token @@@@ NEL.to_list tokens)\n";
```
nberth, line 153: There's an `NEL.pp` already ;-)

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -0,0 +1,421 @@
...
+  Fmt.pf ppf "type completion_entry =\n";
+  Fmt.pf ppf "  | K of token NEL.t\n";
+  let custom_types = Nonterminal.fold (fun nonterm acc ->
+    match nonterminal_filter_map nonterm with
+    | Some (Custom s) -> (Fmt.pf ppf "  | %s\n" s; s::acc)
+    | _ -> acc) [] in
+  emit_pp_completion_entry ppf custom_types;
+  emit_complentryset ppf custom_types
+
+let nullable_nonterminals = Lr1.tabulate begin fun lr1 ->
```
nberth, line 184:
```suggestion
let nullable_nonterminals: lr1 -> SymbolSet.t =
  Lr1.tabulate begin fun lr1 ->
```

nberth: Why not sets of non-terminals?

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -0,0 +1,421 @@
...
+    Fmt.(using Lr1.to_int int)
+    Fmt.(pp_brackets @@ pp_list @@ (using Production.to_int int))
+    "[]"
+
+let completion_entries_of =
+  let rec continue_while_terminal rhs i l =
+    match rhs.(i) with
+        | T t, _, _ when is_valid_for_comp t ->
+            continue_while_terminal rhs (i+1) (t::l)
+        | _ | exception Invalid_argument _ -> List.rev l
```
nberth, line 241: In general, if at least one case expression starts on its own line, please break lines right after `->` (and before any `when`), even if that leads to short lines.

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -0,0 +1,421 @@
...
+    | [] -> acc
+    | defaults -> ProdMap.add_to_list defaults lr1 acc)
+  ProdMap.empty
+  |> inv (ProdMap.fold (fun s l acc -> (l, s)::acc)) []
+  |> pp_match_cases ppf
+    Fmt.(using Lr1.to_int int)
+    Fmt.(pp_brackets @@ pp_list @@ (using Production.to_int int))
+    "[]"
+
+let completion_entries_of =
```
nberth, line 236: Please add the type annotations: here it's `lr1 -> ?`?

nberth (commented):
I think the completion logic that's implemented is pretty cool. Bravo!

Just a few cosmetic comments, questions, and suggestions.

src/lsp/cobol_parser/grammar_tokens.mly
nberth: Oh. It was probably missing from that list.

src/lsp/cobol_parser/keywords/gen_keywords.ml
nberth: I see. `combined_name` maybe (replacing underscores is not really a "cobolization" in the sense of `cobolize` above).

nberth:
Additional: for clarity, it may be worth adding a `grammar_expect.mli`.

src/lsp/cobol_lsp/lsp_completion.ml
```diff
@@ -13,28 +13,33 @@
...
-      else Visitor.skip_children
-
-    method! fold_name name acc =
-      Visitor.skip_children @@ StringSet.add name acc
-
-    end
-  in
-  Cobol_ptree.Visitor.fold_compilation_group visitor ast StringSet.empty
-  |> StringSet.elements
+module Menhir = Cobol_parser.INTERNAL.Grammar.MenhirInterpreter
```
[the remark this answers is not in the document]
nberth: Actually, this is already exported as `Cobol_parser.INSPECT` (from `cobol_engine.ml`).

src/lsp/cobol_parser/cobol_parser.ml
```diff
@@ -21,6 +21,9 @@ module Outputs = Parser_outputs
 module Diagnostics = Parser_diagnostics
 
 module Tokens = Grammar_tokens
+module Recover = Grammar_recover
```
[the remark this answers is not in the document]
NeoKaios:
```ocaml
      let default_value =
        try Some (Cobol_parser.Recover.default_value sym)
        with Not_found -> try Some (Expect.guessed_default_value_of_nullables sym)
        with Not_found -> if debug then  Lsp_io.log_debug "Not found"; None in
```
The default_value for `Menhir.feed`
I guess I can put everything in `guessed_default_value` but that would cause duplicate code

src/lsp/cobol_parser/expect/gen_expect.ml
NeoKaios: Just seemed cleaner than a series of `if ... then ... else`

src/lsp/cobol_parser/cobol_parser.ml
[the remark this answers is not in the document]
nberth: If `sym` is always a nonterminal, I think it's also better to avoid generalizing to a symbol.

src/lsp/cobol_parser/expect/gen_expect.ml
nberth: > Just seemed cleaner than a series of `if ... then ... else`

Sometimes it is. But here it's misleading. If you align them properly (with line breaks (only) right before every `then` and `else`), that is usually readable and avoids the clutter of wildcard patterns.

src/lsp/cobol_parser/expect/gen_expect.ml
NeoKaios: because add_to_list comes from ocaml 5.1 standard lib

nberth: You can add this as a comment near the function definition, but that's not worth adding in the functor name.

src/lsp/cobol_parser/expect/gen_expect.ml
NeoKaios: Yes it holds, I put it because on some cases the guessing the type part fails (ex: type is `int * int option` -> defaults to `None` instead of `(0, None)`). I updated the comment for clarity

It returns a symbol for consistency with `Recover.default_value` and to simplify the logic on the `lsp_completion` side of things

nberth: I believe `symbol` is defines as `type symbol = T of terminal | N of nonterminal`. So passing `(N nt)` to `Recover.default_value` is not an issue. We can postpone that change of API for later if you prefer.

src/lsp/cobol_parser/grammar_expect.mli
```diff
@@ -0,0 +1,32 @@
...
+open Grammar.MenhirInterpreter
+
+type completion_entry =
+  | K of token Cobol_common.Basics.NEL.t
+  | QualifiedRef
+  | ProcedureRef
+
+val pp_completion_entry: completion_entry Fmt.t
+
+module CompEntrySet : (Set.S with type elt = completion_entry)
```
nberth, line 23: Do you really need to expose that module? I suggest to simply expose `compare_completion_entry` (and maybe export a module `Completion_entry: Set.OrderedType with type t = completion_entry`), and then use `Set.Make` whenever necessary. That way other structures (maps, etc) can be defined.

src/lsp/cobol_parser/package.toml
```diff
@@ -112,7 +112,9 @@ dune-trailer = """
   (action
     (with-stdout-to %{targets}
        (run %{exe:./expect/gen_expect.exe} %{deps}
-            --external-tokens Grammar_tokens))))
+            --external-tokens Grammar_tokens
+            --recover-module Grammar_recover
```
nberth, line 116: From the point of view of `expect`, this is `defaults`, not `recover`.

src/lsp/cobol_parser/parser_engine.mli
```diff
@@ -119,12 +119,10 @@ val rewind_and_parse
 
 (** {1 Rewinding for inspection} *)
 
-module INSPECT: module type of Grammar.MenhirInterpreter
```
nberth, line 122: Please leave that one as is (it's internal and makes the type definition below a bit clearer).

src/lsp/cobol_parser/grammar_expect.mli
```diff
@@ -0,0 +1,32 @@
...
+
+val reducible_productions_in: env:_ env -> production list
+
+val acceptable_nullable_nonterminals_in: env:_ env -> xsymbol list
+
+val acceptable_terminals_in: env:_ env -> completion_entry list
```
nberth, line 29: Here there's a mismatch between the name and the type. That one should be called `completion_entries_in`. "acceptable" is not needed as that's implicit in the module name ("Expect").

src/lsp/cobol_parser/grammar_expect.mli
```diff
@@ -0,0 +1,32 @@
...
+
+val guessed_default_value_of_nullables: 'a nonterminal -> 'a
```
nberth, line 31:
```suggestion
val default_value_of_nullable: 'a nonterminal -> 'a
```
(Does this assume the given non-terminal must be nullable?)

src/lsp/cobol_parser/grammar_expect.mli
nberth: (Same above). Now I think all of this may end up being called `grammar_inspect.ml` and be bound as `Cobol_parser.Inspect`… I'll see later.

src/lsp/cobol_lsp/lsp_completion.ml
```diff
@@ -97,73 +54,69 @@ let range (pos:Position.t) text =
...
+  let pp ppf =
+    let rec inner ?(space_before=true) = function
+      | [] -> ()
+      | hd::tl ->
+        let token' = hd &@ Srcloc.dummy in
+        if hd <> Cobol_parser.Tokens.PERIOD && space_before then Fmt.sp ppf ();
+        Cobol_parser.INTERNAL.pp_token ppf token';
+        inner tl
+    in inner ~space_before:false
+  in Pretty.to_string "%a" pp (Basics.NEL.to_list tokens)
```
nberth, line 90:
```suggestion
  in
  Pretty.to_string "%a" pp (Basics.NEL.to_list tokens)
```
Maybe that would be clearer with a match on `PERIOD` instead of an equality test on `hd`.

nberth: Please add a comment "TODO: Cobol_parser.Tokens.pp".

src/lsp/cobol_lsp/lsp_completion.ml
```diff
@@ -97,73 +54,69 @@ let range (pos:Position.t) text =
...
-    if i == 0 then (h, t) else
-    match t with
-    | [] -> (h, [])
-    | hd::tl -> inner (hd::h) tl (i-1)
-  in let h, t = inner [] l i in List.rev h, t
-
-let pp_state ppf state = (* for debug *)
-  Fmt.pf ppf "%d" (Menhir.number state)
+  let pos = range.end_ in
+  List.flatten @@ List.map (function
```
nberth, line 94:
```suggestion
  List.flatten @@ List.rev_map (function
```
(if order is irrelevant at this point)

Not sure what `to_completion_item` does. It probably needs a better name.

src/lsp/cobol_lsp/lsp_completion.ml
```diff
@@ -97,73 +54,69 @@ let range (pos:Position.t) text =
...
+          ~kind:CompletionItemKind.Variable ~range
+          (qualname_proposals ~filename pos group)
+      | ProcedureRef ->
+        to_completion_item
+          ~kind:CompletionItemKind.Module ~range
+          (procedure_proposals ~filename range.end_ group)
+      | K tokens -> begin
+          try let tokens_string = string_of_K tokens in
+            [completion_item ~kind:CompletionItemKind.Keyword ~range tokens_string]
+          with Not_found -> [] end)
```
nberth, line 106:
```suggestion
      | K tokens ->
          try [ completion_item ~kind:CompletionItemKind.Keyword ~range @@
                string_of_K tokens ]
          with Not_found -> [])
```

src/lsp/cobol_lsp/lsp_completion.ml
```diff
@@ -97,73 +54,69 @@ let range (pos:Position.t) text =
...
-    | hd::tl -> inner (hd::h) tl (i-1)
-  in let h, t = inner [] l i in List.rev h, t
-
-let pp_state ppf state = (* for debug *)
-  Fmt.pf ppf "%d" (Menhir.number state)
+  let pos = range.end_ in
+  List.flatten @@ List.map (function
+      | Expect.QualifiedRef ->
+        to_completion_item
+          ~kind:CompletionItemKind.Variable ~range
```
nberth, line 97: Not even sure you need the `CompletionItemKind` qualifiers at this point.

src/lsp/cobol_lsp/lsp_completion.ml
```diff
@@ -173,25 +126,25 @@ let expected_tokens base_env =
       | None -> snd (Srcloc.as_lexloc Srcloc.dummy)
       | Some Menhir.Element (_, _, _, pos) -> pos in
     if debug then (Lsp_io.log_debug "In State: %a" pp_env env; let tok = Expect.acceptable_terminals_in ~env in if List.length tok > 0 then Lsp_io.log_debug "Gained %d entries [%a]" (List.length tok) Fmt.(list ~sep:(any ";") Expect.pp_completion_entry) tok);
-    let productions = Expect.reducable_productions_in ~env in
```
nberth: All that `expected_tokens` is a bit messy. But we can leave it as is and anybody wanting to understand will refactor a bit first (that's typically what happens in OCaml: refactoring helps understanding).

src/lsp/cobol_lsp/lsp_completion.ml
```diff
@@ -173,25 +126,25 @@ let expected_tokens base_env =
...
-      Option.fold ~none:acc ~some:(fun a ->
-        let new_env = Menhir.feed sym pos a pos env in
-        if debug then Lsp_io.log_debug "NULLABLES: On %a\nGot to %a" pp_env_stack env pp_env_stack new_env;
-        inner new_env acc) default_value)
-    acc nullables in
+        (List.to_seq @@ Expect.acceptable_terminals_in ~env) acc in
+    let acc = List.fold_left (fun acc -> function
+        | Menhir.X T _ -> acc
+        | X N nt ->
+          let default_value = try Some (Expect.guessed_default_value_of_nullables nt)
```
nberth, line 136: Here we need some newlines.

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -105,55 +111,57 @@ module SuperSet (OrderedType : Set.OrderedType) = struct
...
+  end)
 
-module SymbolMap = Map5(struct
-  type t = SymbolSet.t
-  let compare = SymbolSet.compare
-end)
+module NonterminalMap = Make_map(struct
+    type t = NonterminalSet.t
+    let compare = NonterminalSet.compare
+  end)
```
nberth, line 153: I did not spot: that's more than a typical map. ~~`NonterminalMap`~~/`Nonterminals_mapping` (so the `s` stands a little).

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -167,234 +175,238 @@ module CompEntrySet = Set.Make(struct
...
+    ProdMap.empty
   |> inv (ProdMap.fold (fun s l acc -> (l, s)::acc)) []
   |> pp_match_cases ppf
     Fmt.(using Lr1.to_int int)
     Fmt.(pp_brackets @@ pp_list @@ (using Production.to_int int))
     "[]"
 
-let completion_entries_of =
+let completion_entries_of: lr1 -> CompEntrySet.t =
   let rec continue_while_terminal rhs i l =
```
nberth, line 246: Maybe that should be named `accumulate_terminals` instead.

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -167,234 +175,238 @@ module CompEntrySet = Set.Make(struct
...
-  EzString.starts_with ~prefix:"loption_" mangled_name ||
-  EzString.starts_with ~prefix:"list_" mangled_name
-  -> Some "[]"
-  | _ when EzString.ends_with ~suffix:"unit" typ
-  -> Some "()"
-  | _ when EzString.ends_with ~suffix:"bool" typ ||
-  EzString.starts_with ~prefix:"boption_" mangled_name
-  -> Some "false"
-  | _ -> None
+  if EzString.ends_with ~suffix:" option" typ ||
```
nberth, line 295: So here you don't deal with `int * int option`. Where is this handled?

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -167,234 +175,238 @@ module CompEntrySet = Set.Make(struct
...
+  else if String.equal typ "bool" ||
+          EzString.starts_with ~prefix:"boption_" mangled_name
+  then Some "false"
+  else None
 
 let is_nullable_without_recovery = Nonterminal.tabulate begin fun nt ->
-  Nonterminal.nullable nt &&
+    Nonterminal.nullable nt &&
     Nonterminal.attributes nt
     |> List.find_opt (Attribute.has_label "recovery")
```
nberth, line 313: Do you really need to deal with `recovery` attributes here? Wouldn't this be given by `default_value` anyways?

src/lsp/cobol_parser/grammar_expect.mli
NeoKaios: Not necessarily but since I only need the default for nullable, many non-nullable nonterminal will raise Not_found (except those with `@recovery`)

src/lsp/cobol_lsp/lsp_completion.ml
```diff
@@ -13,91 +13,148 @@
...
-      else Visitor.skip_children
-
-    method! fold_name name acc =
-      Visitor.skip_children @@ StringSet.add name acc
-
-    end
+module Menhir = Cobol_parser.Grammar_interpr
+module Expect = Cobol_parser.Expect
+
+let qualname_proposals ~filename pos group =
```
emilienlemaire, line 24: I think this should be `qualnames_proposal` (it is a proposal of multiple qualnames, if I understand it right)

src/lsp/cobol_lsp/lsp_completion.ml
```diff
@@ -13,91 +13,148 @@
...
+  match Lsp_lookup.cobol_unit_at_position ~filename pos group with
+  | None -> []
+  | Some cu ->
+    List.filter_map Cobol_data.Item.def_qualname cu.unit_data.data_items.list
+
+let procedure_proposals ~filename pos group =
```
emilienlemaire, line 30: ditto with `procedures`

src/lsp/cobol_lsp/lsp_completion.ml
```diff
@@ -13,91 +13,148 @@
...
+  Fmt.pf ppf "TOP(%a)  STACK: %a" pp_env env
+    (Fmt.list ~sep:(Fmt.any " ") pp_env) (get_stack env)
+
+let debug = false
+let expected_tokens base_env =
+  let rec inner env acc =
+    let pos = match Menhir.top env with
+      | None -> snd (Srcloc.as_lexloc Srcloc.dummy)
+      | Some Menhir.Element (_, _, _, pos) -> pos in
+    if debug then (Lsp_io.log_debug "In State: %a" pp_env env; let tok = Expect.acceptable_terminals_in ~env in if List.length tok > 0 then Lsp_io.log_debug "Gained %d entries [%a]" (List.length tok) Fmt.(list ~sep:(any ";") Expect.pp_completion_entry) tok);
```
emilienlemaire, line 128:
```suggestion
    if debug
    then (Lsp_io.log_debug "In State: %a" pp_env env;
    	let tok = Expect.acceptable_terminals_in ~env in
    	if List.length tok > 0
    	then Lsp_io.log_debug
    		"Gained %d entries [%a]" (List.length tok)
    		Fmt.(list ~sep:(any ";") Expect.pp_completion_entry) tok);
```

emilienlemaire: (edit tabs into spaces)

src/lsp/cobol_lsp/lsp_completion.ml
NeoKaios: I think it's useful for the user, it shows as a little icon in the completion menu on vsc

emilienlemaire (commented):
Overall very good, just a few formatting notes:
* Lines should avoid being longer than 80 chars and never be longer than 100 (`:set colorcolumn="80"` in vim (or `vim.opt.colorcolumn` in lua))
* A new line should almost always follow a `;` and a `in`, and be present before most keywords (`if` `then` `else` `let`)
* Don't be afraid of small lines, be deadly scared of lines that are too long :scream:

src/lsp/cobol_lsp/lsp_completion.ml
NeoKaios: The base for the sanitizer in `DEBUG.emit_nullable_unrecoverable`

src/lsp/cobol_parser/expect/gen_expect.ml
NeoKaios: I added a `@default`, it was the only case in the grammar with a product type ending with option:
```ocaml
from_to_characters_opt [@default (None, None)]:
```

src/lsp/cobol_lsp/lsp_completion.ml
nberth: That's not what I meant (maybe). I meant you can just remove the "`CompletionItemKind.`" in that code, and that will read as `to_completion_item ~kind:Variable ~range …`.

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -167,234 +175,238 @@ module CompEntrySet = Set.Make(struct
...
-    |> List.find_opt (Attribute.has_label "default")
-    |> Option.map Attribute.payload in
-  match given_default with
-    | Some _ -> given_default
-    | None -> guessed_default
+    if not (is_nullable_without_recovery nt)
+    then None
+    else
+      let mangled_name = Nonterminal.mangled_name nt in
+      let guessed_default = Option.bind
```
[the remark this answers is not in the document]
NeoKaios: Not sure I understand what you mean

src/lsp/cobol_parser/grammar_expect.mli
nberth: Maybe that should be renamed into `default_nonterminal_value` then.

src/lsp/cobol_parser/expect/gen_expect.ml
NeoKaios: Well yes but no need to create an entry with a guessed type in the function if we know the case will be caught by the recovery, this would only add to the risk of wrongly guessing the type...

src/lsp/cobol_parser/expect/gen_expect.ml
nberth: I'm curious as to what happens if you remove that `@default` annotation. Does the `gen_expect.ml` script fail? or the compilation of `grammar_expect.ml`? a potential error at runtime (upon a completion request)? or just some completion cases that don't show up?

src/lsp/cobol_parser/expect/gen_expect.ml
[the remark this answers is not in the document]
nberth: The binding to `guessed_default` (and maybe `mangled_name`) is not used in case the `match` below returns `Some _`. So you can add the `Option.bind …` directly in the `None` case.

src/lsp/cobol_parser/expect/gen_expect.ml
nberth: I think that should be the other way round: the function that guesses the type should be a sort of under-approximation, i.e, return a result only if it is sure that is correct. The way to do that is probably by working only on the type string, and checking some basic regexps. For instance if the type string matches  `[^ ]+ option`, then it is surely a `None`. That's ok if less expressive cases require a `@default` attribute.

Apologies if I lead you the wrong way with checks for suffixes. Here you may just `String.split_on_char ' '` and check the last element is `"option"` or `"list"`. Or use regexps. That's up to you.

With a sound guess, we can leave the remaining cases to the `Defaults` module (effectively the recovery).

src/lsp/cobol_parser/expect/gen_expect.ml
NeoKaios: In this case the `grammar_expect.ml` fails to compile because some `match` case is misstyped:
```shell
File "src/lsp/cobol_parser/grammar_expect.ml", line 3430, characters 32-36:
3430 |   | N_from_to_characters_opt -> None
                                       ^^^^
Error: This expression should not be a constructor, the expected type is a
```

nberth: Ok, good to known it's a static check.

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -289,16 +294,23 @@ let guess_default_value ~mangled_name ~typ =
...
+let extra_default = match !extra_default with
+| "" -> None
+| s -> Some s
+
 let best_guess_default_value = Nonterminal.tabulate begin fun nt ->
     if not (Nonterminal.nullable nt)
     then None
     else
-      let given_default =
+      let given_or_extra_default =
```
nberth, line 305: `given_default` was totally fine.

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -312,16 +324,30 @@ let default_nonterminal_value (type a): a Menhir.nonterminal -> a = function
...
+let ensure_all_nullable_nonterminal_have_default () =
+  let nt_without_default = Nonterminal.fold (fun nt acc -> begin
+        if Nonterminal.nullable nt
+        && best_guess_default_value nt == None
+        then nt::acc else acc
+      end) [] in
+  match List.hd nt_without_default with
+  | nt ->
+      let error_msg = Pretty.to_string
+        "Unable to find default value for the nonterminal '%s', please provide a @default annotation"
```
nberth, line 346:
```suggestion
        "Unable to find a default value for nonterminal '%s'; please provide a @@default annotation"
```

nberth: You can also `Pretty.failwith "…"` directly (no need for the extra binding)

nberth: Should also probably all go in a `Nonterminal.iter begin fun nt -> …`. That'd be much cleaner than with the preliminary `fold`. (cf other comment)

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -312,16 +324,30 @@ let default_nonterminal_value (type a): a Menhir.nonterminal -> a = function
...
+    match best_guess_default_value nt with
+    | None -> acc
+    | Some default -> ([nt], default)::acc
+  end []
   |> pp_match_cases ppf
     Fmt.(using Nonterminal.mangled_name (any "N_" ++ string))
     Fmt.string
     "raise Not_found"
 
+let ensure_all_nullable_nonterminal_have_default () =
```
nberth, line 337:
```ocaml
let __ensure_every_nullable_nonterminal_has_default_value =
  Nonterminal.iter begin fun nt -> …
```
So you don't need to evaluate it below.

nberth (approved):

PR #272, opened 2024-04-22
SQL parser

LilyOlivier:
draft pull request for David's rewiews

PR #308, opened 2024-06-26
Fix: `drom project -f; make opam-cross`

nberth:
Followup of #272: update project and cross-compilation files.


opam/sql_parser.opam
```diff
@@ -55,7 +55,7 @@ depends: [
   "cobol_parser" {= version}
   "cobol_common" {= version}
   "ppx_deriving" {>= "5.2.1"}
-  "menhir" {= "20230415"}
+  "menhir" {>= "20230415"}
```
emilienlemaire, line 58: There seem to be a mismatch with `src/lsp/sql_parser/package.toml:71`

src/lsp/sql_parser/package.toml
```diff
@@ -68,7 +68,7 @@ version = ">=1.2"
 # package tools dependencies
 [tools]
 ppx_deriving = ">=5.2.1"
-menhir = ">=20230415"
+menhir = "20230415"
```
emilienlemaire, line 71: ditto

emilienlemaire (approved):
Ok to merge after checking my concern

ddeclerck (approved):

src/lsp/sql_parser/package.toml
nberth: A version without a relational operator in a `package.toml` is interpreted as a lower-bound by `drom`. The `>=` is optional (but I can put it back anyways).

PR #310, opened 2024-07-03
Fix object-view syntax

nberth:


ddeclerck (approved):
LGTM

test/syntax-todo/syntax-todo.at
```diff
@@ -863,3 +863,19 @@ AT_CHECK([$SUPERBOL check --recovery=false --std=mf prog.cob], [0],
...
+# Note: this program's semantics makes no sense; it is used to test
+# proper parsing of `AS` in object views.
+AT_DATA([prog.cob],[
+program-id. prog.
+procedure division.
+    cancel a as universal.
+])
+AT_CHECK([$SUPERBOL check syntax --recovery=false --std=default prog.cob], [0], [Checking `prog.cob'
+])
+AT_CLEANUP
```
emilienlemaire, line 881: Does this test fail? If not, then it must be moved to `syntax-done.at`, if yes, then you should promote it so the failure is in `AT_CHECK`, and we can move it to `syntax-done.at` once we handle this syntax.

emilienlemaire (approved):
Ok to merge after modifications to the test are done :+1:

test/syntax-todo/syntax-todo.at
nberth: Actually this was done using two commits: one that adds the test in `syntax-todo.at` (with expected failure), the other bringing the fix and promoting the test. As I said at some point I don't like having separate files for that as it almost renders auto-promotion useless. One `syntax.at` may be enough, possibly split into other kinds of sub-categories (maybe in another PR).

emilienlemaire: Yes, but until this other PR is done, we should keep to the current way, to avoid more confusion about those files.

PR #312, opened 2024-07-04
Add some QOL feature to context-completion 

NeoKaios:
- eager completion: suggest to complete more than one token when there is a series of mandatory tokens.
- Added newline after `.`
- Added support for intrinsic function
- Added automatic case detection


src/lsp/cobol_common/basics.ml
```diff
@@ -52,13 +52,34 @@ module NEL = struct
...
+      | One a, One b -> cmp a b
+      | a :: _, One b  ->
+        let c = cmp a b in
+        if c = 0 then 1 else c
+      | One a, b :: _ ->
+        let c = cmp a b in
+        if c = 0 then -1 else c
+      | a :: a', b :: b' ->
+        let c = cmp a b in
+        if c = 0 then aux a' b' else c
```
nberth, line 82: I've had a look at the stdlib, and their `List.compare` is the lexicographic one. So I'd be in favor of using `compare` for that one, and maybe `compare_whole` for the previous one (I don't think it's used a lot so there might not be so many renaming to do in the whole project).

nberth: I the documentations, just use "compares" instead of "sorts".

src/lsp/cobol_parser/expect/gen_expect.ml
```diff
@@ -274,6 +282,215 @@ let completion_entries_in ~env : Completion_entry.t list =
...
+end
+
+let nonterminal_pre_comp_filter_map: nonterminal -> pre_comp option = fun nonterm ->
+  Nonterminal.attributes nonterm |>
+  List.find_opt (Attribute.has_label "completion") |>
+  Option.map (fun attr -> UserDefinedWords (Attribute.payload attr))
+
+let should_log stack =
+  let lr1 = NEL.last stack in
+  Lr1.to_int lr1 == -1
```
nberth, line 316: This is a bit opaque…

test/lsp/lsp_completion.ml
nberth, line 1: Given the size of these tests, I'd be in favor or splitting it into `lsp_basic_completion.ml` (almost identical copy of previous `lsp_completion.ml`) and `lsp_eager_completion.ml` (the second can reuse `Lsp_basic_completion.completion_positions`).

test/lsp/lsp_completion.ml
```diff
@@ -76,7 +79,7 @@ let%expect_test "division-and-section-completion" =
        3           PROGRAM-ID . prog.
        4           DATA DIVISION.
     (line 1, character 24):
-    List of completions (1 entries): [DIVISION.]
+    List of completions (1 entries): [DIVISION.\n]
```
nberth, line 82: This is the only occurrence of a completion from the middle of a word that I saw. In general what is the expected behaviors in these cases?

NeoKaios: The completion should start at the beginning of the word word (this is computed with the `range` function in `lsp_completion`.
I will add a test for that

test/lsp/lsp_completion.ml
NeoKaios: My goal was to have the same cobol code tested on both eager and non-eager completion... Should I also reuse the cobol programs ?

nberth: In that case, I'd be to expand the `completion_positions` utility to provide both results side-by-side via two distinct calls to the completion function.

test/lsp/lsp_completion.ml
```diff
@@ -15,10 +15,14 @@ open EzCompat                                                    (* StringMap *)
 open Lsp.Types
 open Lsp_testing
 
+let map_to_string items =
+  List.map begin fun (item: CompletionItem.t) ->
```
nberth, line 19: Fmt.fmt "%S" (minus first and last chars that are quotes).

test/lsp/lsp_completion.ml
```diff
@@ -30,26 +34,154 @@ let completion_positions (doc, positions) : string -> unit =
...
     | Some `CompletionList { items; _ } when items == [] ->
-        Pretty.out "Empty completion list@."
+      Pretty.out "Empty eager-completion list@."
     | Some `CompletionList { items; _ } ->
-        let pp_comp_item ppf (item: CompletionItem.t) =
-          Fmt.string ppf item.label in
-        Pretty.out "List of completions (%d entries): [%a]\n"
-          (List.length items) (Fmt.list ~sep:(Fmt.any ";") pp_comp_item) items;
-
+      Pretty.out "List of eager-completion (%d entries): [%a]\n"
```
nberth, line 59: `"@[<hov 2>Eager (%d entries):@;[%a]@\n"`

and then `sep:Fmt.semicolon` below.

test/lsp/lsp_completion.ml
nberth, line 1: Looks like the lists are always in reversed lexicographical order. Is this intended?

NeoKaios: It's not intended but it's expected, it's due to the `List.rev_map` [here](https://github.com/OCamlPro/superbol-studio-oss/pull/312/files#diff-4860399da4d5d6d03f7b2c3aa54a2498c2b72a7bd81fb5b25e6b51bd8a1016a2R139)

nberth: Ah they are already sorted due to `.elements` from the set. In this case it would be better to use `EzList.tail_map` instead (so the question does not arise again).

nberth:
@NeoKaios Could you just add a link to this PR in the CHANGELOG? Then we'll merge.

nberth (approved):
Thanks :tada: !

PR #314, opened 2024-07-11
Refactor `Expect` api for completion

NeoKaios:
Replaced `reducible_productions_in` and `nullable_nonterminals_in` with `actions_in`
Created new type `action` to unify both `feed` and `reduce` action possibility


nberth (approved):
Looks fine, thanks!

PR #321, opened 2024-07-24
Bump `lsp` library to version `1.18.0`

NeoKaios:


src/lsp/cobol_lsp/package.toml
```diff
@@ -62,7 +62,7 @@ cobol_typeck = "version"
 superbol_preprocs = "version"
 superbol_project = "version"
 jsonrpc = ">=1.15"
-lsp = ">=1.15 <1.16"
+lsp = "=1.18.0"
```
nberth, line 65:
```suggestion
lsp = "1.18"
```
(This implicitly means "≥1.18"; should be enough)

nberth (commented):
I'll see to add `lsp` to the cross-compilation repositories.

nberth (approved):

PR #331, opened 2024-07-26
Relax order of paragraphs in `CONFIGURATION SECTION`

### LINKED ISSUE #269: Relax order of paragraphs in the `ENVIRONMENT DIVISION`
The only other problems concerning world cities are the `>>D` compiler directive and the order of the paragraphs in the environment division (worldcities uses `repository` before `special-names` which is not allowed by our parser).

_Originally posted by @emilienlemaire in https://github.com/OCamlPro/superbol-studio-oss/issues/267#issuecomment-2066729585_

NeoKaios:
Closes #269

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,8 @@
 - COBOL language configuration for highlighting matching brackets and auto-insertion of line numbers in fixed-format code [#330](https://github.com/OCamlPro/superbol-studio-oss/pull/330)
 
 ### Fixed
-- Word wrapping in presence of hyphens [#330](https://github.com/OCamlPro/superbol-studio-oss/pull/330)
 - Improvements to the grammar [#331](https://github.com/OCamlPro/superbol-studio-oss/pull/331)
+- Word wrapping in presence of hyphens [#330](https://github.com/OCamlPro/superbol-studio-oss/pull/330)
 
 
 ## [0.1.3] Fourth α release (2024-07-24)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Fixed
 - Word wrapping in presence of hyphens [#330](https://github.com/OCamlPro/superbol-studio-oss/pull/330)
+- Improvements to the grammar [#331](https://github.com/OCamlPro/superbol-studio-oss/pull/331)
 
 
 ## [0.1.3] Fourth α release (2024-07-24)
--- a/src/lsp/cobol_parser/grammar.mly
+++ b/src/lsp/cobol_parser/grammar.mly
@@ -623,16 +623,41 @@ let environment_division :=
 
 (* ------------- ENVIRONMENT DIVISION / CONFIGURATION SECTION -------------- *)
 
+let unorder2_opt [@recovery (None, None, AB)] (A, B) :=
+ | a = A; b = ro(B); { (Some a, b, AB) }
+ | b = B; a = ro(A); { (a, Some b, BA) }
+ | { (None, None, AB) }
+
+let unorder3_opt [@recovery (None, None, None, (A3, AB))] (A, B, C) :=
+ | a = A; u = unorder2_opt(B, C); { prepend_as_trio @@ `A (Some a, u) }
+ | b = B; u = unorder2_opt(A, C); { prepend_as_trio @@ `B (Some b, u) }
+ | c = C; u = unorder2_opt(A, B); { prepend_as_trio @@ `C (Some c, u) }
+ | { (None, None, None, (A3, AB)) }
+
+let unorder4_opt [@recovery (None, None, None, None, (A4, (A3, AB)))]
+                 (A, B, C, D) :=
+ | a = A; u = unorder3_opt(B, C, D); { prepend_as_quartet @@ `A (Some a, u) }
+ | b = B; u = unorder3_opt(A, C, D); { prepend_as_quartet @@ `B (Some b, u) }
+ | c = C; u = unorder3_opt(A, B, D); { prepend_as_quartet @@ `C (Some c, u) }
+ | d = D; u = unorder3_opt(A, B, C); { prepend_as_quartet @@ `D (Some d, u) }
+ | { (None, None, None, None, (A4, (A3, AB))) }
+
 let configuration_section :=
- | CONFIGURATION; SECTION; "."; body = rl(configuration_body);
-   { build_configuration_section body }
-
-let configuration_body ==
- | p = loc(source_computer_paragraph); { P_SOURCE_COMPUTER_PARAGRAPH p }
- | p = loc(object_computer_paragraph); { P_OBJECT_COMPUTER_PARAGRAPH p }
- | p = loc(special_names_paragraph);   { P_SPECIAL_NAMES_PARAGRAPH p }
- | p = loc(repository_paragraph);      { P_REPOSITORY_PARAGRAPH p } (* +COB2002 *)
-
+ | CONFIGURATION; SECTION; ".";
+   unorder = unorder4_opt(loc(source_computer_paragraph),
+                          loc(object_computer_paragraph),
+                          loc(special_names_paragraph),
+                          loc(repository_paragraph));
+  { let (source_computer_paragraph,
+         object_computer_paragraph,
+         special_names_paragraph,
+         repository_paragraph,
+         order) = unorder in
+    { source_computer_paragraph;
+      object_computer_paragraph;
+      special_names_paragraph;
+      repository_paragraph;
+      order } }
 
 (* ENVIRONMENT DIVISION / CONFIGURATION SECTION / SOURCE-COMPUTER PARAGRAPH *)
 
--- a/src/lsp/cobol_parser/grammar.mly
+++ b/src/lsp/cobol_parser/grammar.mly
@@ -624,16 +624,14 @@ let environment_division :=
 (* ------------- ENVIRONMENT DIVISION / CONFIGURATION SECTION -------------- *)
 
 let configuration_section :=
- | CONFIGURATION; SECTION; ".";
-   source_computer_paragraph = ro(loc(source_computer_paragraph));
-   object_computer_paragraph = ro(loc(object_computer_paragraph));
-   special_names_paragraph   = ro(loc(special_names_paragraph));
-   repository_paragraph      = ro(loc(repository_paragraph));     (* +COB2002 *)
-   { { source_computer_paragraph;
-       object_computer_paragraph;
-       special_names_paragraph;
-       repository_paragraph } }
+ | CONFIGURATION; SECTION; "."; body = rl(configuration_body);
+   { build_configuration_section body }
 
+let configuration_body ==
+ | p = loc(source_computer_paragraph); { P_SOURCE_COMPUTER_PARAGRAPH p }
+ | p = loc(object_computer_paragraph); { P_OBJECT_COMPUTER_PARAGRAPH p }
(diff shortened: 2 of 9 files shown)


NeoKaios:
Not sure this is the correct way of doing it....

nberth:
> Not sure this is the correct way of doing it....

Thankfully the combinatorics would be on your side with this approach. However, I'd rather keep the order of declarations consistent between the input and the parse-tree. So relaxing the order should be done by altering the parse-tree types, using a list of paragraphs (and relying on an additional check for duplicates, maybe left for later). We do something similar for the sections in the `DATA DIVISION`.

GitMensch:
Note: this is only part of the issue, MF (and likely others) also allow `INPUT-OUTPUT` and `CONFIGURATION` sections being in wrong order, which we just recently [added to GnuCOBOL](https://sourceforge.net/p/gnucobol/code/5304/) - you likely want to relax the order there in this PR as well

src/lsp/cobol_parser/grammar_utils.ml
```diff
@@ -94,3 +94,29 @@ let build_data_division = function
...
+    let section = build_configuration_section tl in
+    match hd with
+    | P_SOURCE_COMPUTER_PARAGRAPH p ->
+      { section with source_computer_paragraph = Some p }
+    | P_OBJECT_COMPUTER_PARAGRAPH p ->
+      { section with object_computer_paragraph = Some p }
+    | P_SPECIAL_NAMES_PARAGRAPH p ->
+      { section with special_names_paragraph = Some p }
+    | P_REPOSITORY_PARAGRAPH p ->
+      { section with repository_paragraph = Some p }
```
nberth, line 121: Here you take only the last right? (I'm having second thoughts on what I said about keeping the order in the parse-tree… — maybe getting a "stricter" grammar is better than keeping that property, since that allows for more precise completion suggestions…)

src/lsp/cobol_parser/grammar.mly
```diff
@@ -623,18 +623,76 @@ let environment_division :=
...
+ | d = D; u = unorder3_opt(A, B, C); { let (a, b, c, o) = u in (a, b, c, Some d, from_trio `ABC o) }
+ | { (None, None, None, None, ABCD) }
+ *)
+
+let unorder2_opt (A, B) ==
+ | a = A; b = ro(B); { (Some a, b, AB) }
+ | b = B; a = ro(A); { (a, Some b, BA) }
+ | { (None, None, AB) }
+let unorder3_opt (A, B, C) ==
+ | a = A; u = unorder2_opt(B, C); { let (b, c, o) = u in (Some a, b, c, from_duo' `BC o) }
```
nberth, line 669: Couldn't we have something that reads `prepend_as_trio `A `BC o` directly instead?
Or even something parametric like:
```suggestion
 | a = A; u = unorder2_opt(B, C); { prepend_as_trio `A u }
```
with the implicit that `u` encodes a duo `BC` thanks to the ``A`?

src/lsp/cobol_ptree/env_division.ml
```diff
@@ -25,12 +25,20 @@ type environment_division =
...
+    special_names_paragraph: special_names_paragraph with_loc option order;
+    repository_paragraph: repository_paragraph with_loc option order;   (* +COB2002 *)
+  } *)
 and configuration_section =
   {
     source_computer_paragraph: source_computer_paragraph with_loc option;
     object_computer_paragraph: object_computer_paragraph with_loc option;
     special_names_paragraph: special_names_paragraph with_loc option;
     repository_paragraph: repository_paragraph with_loc option;   (* +COB2002 *)
+    order: permutation_quartet';
```
nberth, line 41: We should not have a prime here (that'd be confusing). At worse the prime is for temporary stuff used during construction of the parse-tree.

src/lsp/cobol_ptree/common.ml
```diff
@@ -30,3 +30,203 @@ type 'a nel = 'a NEL.t
...
+(* order *)
+
+type 'a order =
+  | P1 of 'a
+  | P2 of 'a
+  | P3 of 'a
+  | P4 of 'a
+[@@deriving ord]
+
+let (~!) = function
```
nberth, line 44: Are those operators useful?

src/lsp/cobol_ptree/common.ml
```diff
@@ -30,3 +30,203 @@ type 'a nel = 'a NEL.t
...
+  | _ -> acc |> fa a |> fb b |> fc c |> fd d
+
+(* splitted *)
+
+type permutation_duo' =
+  | AB
+  | BA
+[@@deriving ord]
+
+type permutation_trio' =
```
nberth, line 176: I'd go with a simpler representation like:
```suggestion
type trio_head = A3 | B3 | C3
and permutation_trio = trio_head * permutation_duo
```

src/lsp/cobol_ptree/common.ml
```diff
@@ -30,3 +30,203 @@ type 'a nel = 'a NEL.t
...
+[@@deriving ord]
+
+type permutation_quartet' =
+  | A4 of permutation_trio'
+  | B4 of permutation_trio'
+  | C4 of permutation_trio'
+  | D4 of permutation_trio'
+[@@deriving ord]
+
+let from_duo' conv o =
```
nberth, line 189: And here you'd code `prepend_to_duo` directly with the parametric tuple `u`.

src/lsp/cobol_parser/grammar.mly
nberth: Also I think for the sake of the automaton size, I'd be in favor to try using non-inlined items.

src/lsp/cobol_ptree/common.ml
NeoKaios: These are only useful if we choose to go with the `'a order` type to define order

src/lsp/cobol_ptree/env_division.ml
NeoKaios: Yes, the prime is there because I had 2 type that where very similar, but from what I can see you prefer the current prime definition of permutation (recursively defined) instead of the fully defined version (without prime)
So I will remove all the unused stuff in the next commit

src/lsp/cobol_ptree/common.ml
```diff
@@ -30,3 +30,61 @@ type 'a nel = 'a NEL.t
 [@@deriving ord]
 
 let pp_nel ppe = NEL.pp ~fsep:"@ " ~fopen:"" ~fclose:"" ppe
+
+type permutation_duo = AB | BA
+[@@deriving ord]
```
nberth, line 35: I think keeping only the last `deriving` is enough for `type … and … and …` declarations (but I may be wrong on that).

src/lsp/cobol_ptree/env_division.ml
```diff
@@ -560,14 +562,15 @@ let pp_special_names_paragraph ppf sncs =
...
-    Fmt.(option (sp ++ pp_with_loc pp_repository_paragraph)) rp
+    { source_computer_paragraph = scp; object_computer_paragraph = ocp;
+      special_names_paragraph = snp; repository_paragraph = rp; conf_sec_order }
+  =
+  Fmt.pf ppf "CONFIGURATION SECTION.";
+  iter_quartet conf_sec_order (scp, ocp, snp, rp) (
+    Fmt.(option (sp ++ pp_with_loc pp_source_computer_paragraph)) ppf,
+    Fmt.(option (sp ++ pp_with_loc pp_object_computer_paragraph)) ppf,
+    Fmt.(option (sp ++ pp_with_loc pp_special_names_paragraph)) ppf,
+    Fmt.(option (sp ++ pp_with_loc pp_repository_paragraph)) ppf)
```
nberth, line 573:
```suggestion
  iter_quartet conf_sec_order (scp, ocp, snp, rp)
    Fmt.(option (sp ++ pp_with_loc pp_source_computer_paragraph) ppf,
         option (sp ++ pp_with_loc pp_object_computer_paragraph) ppf,
         option (sp ++ pp_with_loc pp_special_names_paragraph) ppf,
         option (sp ++ pp_with_loc pp_repository_paragraph) ppf)
```

src/lsp/cobol_ptree/env_division.ml
```diff
@@ -656,8 +659,9 @@ let pp_input_output_section ppf
...
-=
-  Fmt.pf ppf "ENVIRONMENT DIVISION.%a%a"
-    Fmt.(option (sp ++ pp_with_loc pp_configuration_section)) ec
-    Fmt.(option (sp ++ pp_with_loc pp_input_output_section)) eio
+    { env_configuration = ec; env_input_output = eio; env_order }
+  =
+  Fmt.pf ppf "ENVIRONMENT DIVISION.";
+  iter_duo env_order (ec, eio) (
+    Fmt.(option (sp ++ pp_with_loc pp_configuration_section)) ppf,
+    Fmt.(option (sp ++ pp_with_loc pp_input_output_section)) ppf)
```
nberth, line 667:
```suggestion
  iter_duo env_order (ec, eio)
    Fmt.(option (sp ++ pp_with_loc pp_configuration_section) ppf,
         option (sp ++ pp_with_loc pp_input_output_section) ppf)
```

src/lsp/cobol_ptree/env_division_visitor.ml
```diff
@@ -468,9 +470,10 @@ let fold_configuration_section' (v: _ #folder) =
 
 let fold_environment_division (v: _ #folder) =
   handle v#fold_environment_division
-    ~continue:begin fun { env_configuration; env_input_output } x -> x
-      >> fold_option ~fold:fold_configuration_section' v env_configuration
-      >> fold_option ~fold:fold_input_output_section' v env_input_output
+    ~continue:begin fun { env_configuration; env_input_output; env_order } x ->
+      x >> fold_duo env_order (env_configuration, env_input_output) (
+        fold_option ~fold:fold_configuration_section' v,
+        fold_option ~fold:fold_input_output_section' v)
```
nberth, line 476:
```suggestion
    ~continue:begin fun { env_configuration; env_input_output; env_order } x -> x
      >> fold_duo env_order (env_configuration, env_input_output)
        (fold_option ~fold:fold_configuration_section' v,
         fold_option ~fold:fold_input_output_section' v)
```

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ### Fixed
 - Word wrapping in presence of hyphens [#330](https://github.com/OCamlPro/superbol-studio-oss/pull/330)
+- Improvements to the grammar [#331](https://github.com/OCamlPro/superbol-studio-oss/pull/331)
```
nberth, line 12:
```suggestion
- Improvements to the grammar [#331](https://github.com/OCamlPro/superbol-studio-oss/pull/331)
- Word wrapping in presence of hyphens [#330](https://github.com/OCamlPro/superbol-studio-oss/pull/330)
```

src/lsp/cobol_ptree/common.ml
```diff
@@ -30,3 +30,61 @@ type 'a nel = 'a NEL.t
...
+
+and trio_head = A3 | B3 | C3
+[@@deriving ord]
+
+and permutation_trio = trio_head * permutation_duo
+[@@deriving ord]
+
+and quartet_head = A4 | B4 | C4 | D4
+[@@deriving ord]
+
```
nberth, line 45:
```suggestion
type permutation_duo = AB | BA

and trio_head = A3 | B3 | C3

and permutation_trio = trio_head * permutation_duo

and quartet_head = A4 | B4 | C4 | D4

```

PR #337, opened 2024-08-05
Example of displayed value on hover of `NUMERIC-EDITED` variables

### LINKED ISSUE #296: Improve the information shown when hovering data fields of `-EDITED` category
Some variable with Edited picture string, such as:
```cobol
01 VAR PIC -BZZZ,ZZ9.99.
```
Display really long text on hover, in this case:
```
...
NUMERIC-EDITED(digits = 8, scale = 2, with_sign = false, editions = { Data_picture.TYPES.basics = [(Data_picture.TYPES.FixedInsertion { Data_picture.TYPES.fixed_insertion_symbol = -; fixed_insertion_offset = 0 }); (Data_picture.TYPES.SimpleInsertion { Data_picture.TYPES.simple_insertion_symbols = { Data_picture.TYPES.symbol = B; symbol_occurences = 1 }; simple_insertion_offset = 1 }); (Data_picture.TYPES.SimpleInsertion { Data_picture.TYPES.simple_insertion_symbols = { Data_picture.TYPES.symbol = grouping separator; symbol_occurences = 1 }; simple_insertion_offset = 5 }); (Data_picture.TYPES.SpecialInsertion { Data_picture.TYPES.special_insertion_offset = 9; special_insertion_length = 1 })]; floating = None; zerorepl = (Some { Data_picture.TYPES.zero_replacement_symbol = Z; zero_replacement_ranges = [{ Data_picture.TYPES.floating_range_offset = 2; floating_range_length = 3 }; { Data_picture.TYPES.floating_range_offset = 6; floating_range_length = 2 }] }) })
```

This could be improved by the following:

> One could show edition informations in more or less useful ways. Not in general at first (_i.e,_ putting special insertion and zero replacement aside), a series of `_`s, some being replaced with fixed or simple insertion symbols, and only in the case where we have the length of the data, could be nice. Like `- ___,___.__` in the case above (we have the offsets). Not sure that would be helpful, but that may be worth a try.
> The next step would be to show examples with "typical" data (like `- 123,456.12` for instance).
https://github.com/OCamlPro/superbol-studio-oss/pull/293#discussion_r1618433421

NeoKaios:
Closes #296
Note that picture string with `P` and `E` character are not yet implemented


src/lsp/cobol_data/data_picture.ml
```diff
@@ -1289,3 +1281,245 @@ let unit_test ?(config=config) ~expect picture =
...
+      match of_string config pic with
+      | Ok picture ->
+        let example = example_of ~picture value in
+        let error_msg = Pretty.to_string "ERROR: different result (%s, %f) -> '%s' (expected: '%s')" pic value example expected in
+        if String.equal example expected
+        then test tl
+        else failwith error_msg
+      | Error _ ->
+        failwith @@ Pretty.to_string "ERROR: Unable to form picture with picture-string '%s'\n" pic
+  in test unit_tests
```
nberth, line 1525: This should probably go in a `data_picture_interpr.ml` file. Also the tests should be in the testsuite (the ones added by Fabrice should never have been put here).

nberth: Actually this should not be in the same library. At the moment this is LSP-specific, so having that as `cobol_lsp/lsp_picture_interpr.ml` would do.

nberth:
I guess at the moment adding `lsp = ">=1.18 & <1.19"` in the `cobol_lsp/package.toml` should do fine. We upgrading that should be kept for a separate PR.

src/lsp/cobol_data/data_picture.mli
```diff
@@ -40,6 +40,9 @@ module TYPES: sig
     | Z
     | Zero
 
+  val pp_symbol: symbol Fmt.t
+  val pp_symbol_cobolized: symbol Fmt.t
```
nberth, line 44:
```suggestion
  val pp_symbol: symbol Pretty.printer
  val pp_symbol_cobolized: symbol Pretty.printer
```
Let's use the readable version ;-)

src/lsp/cobol_lsp/lsp_data_info_printer.ml
```diff
@@ -25,16 +25,37 @@ let pp_cobol_block: _ Fmt.t -> _ Fmt.t = fun pp ->
...
+
+let pp_example_of ppf (picture: Cobol_data.Picture.t) =
+  match picture.category with
+  | FixedNum { digits; scale; _ } ->
+      let max = max_value digits scale in
+      let max_str =
+      if Float.is_integer max
+      then string_of_int (int_of_float max)
+      else string_of_float max in
+    Fmt.pf ppf "\n\nExample display value:\n- 0: `%s`\n- %s: `%s`"
```
nberth, line 44: Isn't there a better formatting we could find for this, more on a single line, simply separated with commas? Also `Example display value**s**`, as a heading and without `:`.
Next, I'd put the actual rendering first, and then the actual raw value that is rendered in between parentheses.

test/lsp/lsp_hover.ml
```diff
@@ -292,9 +292,7 @@ let%expect_test "hover-typedef-vars" =
     PIC 999 USAGE DISPLAY
     ```
     NUMERIC(digits = 3, scale = 0, with_sign = false)
-    Example display value:
-    - 0: `000`
-    - 123: `123`
+    Example display values [000] (0) [123] (123)
```
nberth, line 295: I think this should be something like
```
#### Example display values
[`000`] (0), [`123`] (123)
```

nberth: Or actually just `*e.g,*` instead of the heading might do.

test/lsp/lsp_hover.ml
```diff
@@ -469,9 +465,7 @@ let%expect_test "hover-typedef-vars-usage" =
     PIC -BZZZ,ZZ9.99 USAGE DISPLAY
     ```
     NUMERIC(digits = 8, scale = 2, with_sign = false)
-    Example display value:
-    - 0: `        0.00`
-    - 123456.78: `  123,456.78`
+    Example display values [&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;0.00] (0) [&nbsp;&nbsp;123,456.78] (123456.78)
```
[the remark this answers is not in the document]
NeoKaios: It was needed even in between ```, but utf8 nbsp character works nicely !

nberth (commented):
Good job!

Just a small question: is `P` really still not supported (it does not appear to be tested in `test/lsp/lsp_picture_interp.ml`)? About `E`, what are the current limitations and what happens on such items?

NeoKaios:
`P` are still not supported:
I think there is an issue/missing info when building the category of vars whose picture string contains `P`s
Because picture string `V99` and `P9` lead to the exact same category `NUMERIC(digits = 2, scale = 2, with_sign = false)`
(Also `PP` is valid in superbol even tho it's not a valid picture string.)
The handling of PIC `9P` and similar should be doable, but needs to be done separately because of the `scale = -1`

I didn't do `E` because I only found 1 example in the docs. Now that I think about it, I suppose that the implementation reflects scientific notation so it can be done, tho without feedback, as `cobc` issue an error when an `E` is present `libcob: error: cob_move_display_to_edited: invalid PIC character E`   

Atm, items with `E` and `P` simply do not have a `e.g` clause

nberth (approved):

PR #346, opened 2024-08-06
Unify `pp_token` of Text_tokenizer with other pp

NeoKaios:
The old `pp_token` of Text_tokenizer didn't have a prime even though it was a pretty printer for `token with_loc`

### DIFF
--- a/src/lsp/cobol_parser/grammar_tokens_printer.ml
+++ b/src/lsp/cobol_parser/grammar_tokens_printer.ml
@@ -77,8 +77,10 @@ let pp_token: token Pretty.printer = fun ppf ->
     | FIXEDLIT (i, sep, d) -> print "FIXED[%s%c%s]" i sep d
     | FLOATLIT (i, sep, d, e) -> print "FLOAT[%s%c%sE%s]" i sep d e
     | INTERVENING_ c -> print "<%c>" c
-    | tok when Text_keywords.is_intrinsic_token tok ->
+    | INTRINSIC_FUNC _ as tok ->
         print "INTRINSIC_FUNC[%a]" pp_token_string tok
+    | tok when Text_keywords.is_known_intrinsic_token tok ->
+        print "SPECIALIZED_INTRINSIC_FUNC[%a]" pp_token_string tok
     | EOF -> string "EOF"
     | t -> pp_token_string ppf t
 
--- a/src/lsp/cobol_parser/grammar_tokens_printer.ml
+++ b/src/lsp/cobol_parser/grammar_tokens_printer.ml
@@ -1,0 +1,96 @@
+(**************************************************************************)
+(*                                                                        *)
+(*                        SuperBOL OSS Studio                             *)
+(*                                                                        *)
+(*  Copyright (c) 2022-2023 OCamlPro SAS                                  *)
+(*                                                                        *)
+(* All rights reserved.                                                   *)
+(* This source code is licensed under the GNU Affero General Public       *)
+(* License version 3 found in the LICENSE.md file in the root directory   *)
+(* of this source tree.                                                   *)
+(*                                                                        *)
+(**************************************************************************)
+
+open Cobol_common.Srcloc.INFIX
+open Cobol_common.Srcloc.TYPES
+open Grammar_tokens                              (* import token constructors *)
+
+(* --- *)
+
+let combined_tokens =
+  (* /!\ WARNING: None of the constituents of combined tokens may be
+     context-sensitive.
+
+     Rationale: this would considerably complicate retokenization (which is
+     necessary with the current solution to handle context-sensitive
+     keywords) *)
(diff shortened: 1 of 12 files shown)


src/lsp/cobol_parser/text_tokenizer.ml
```diff
@@ -93,40 +92,43 @@ let pp_token_string: Grammar_tokens.token Pretty.printer = fun ppf ->
       try Hashtbl.find combined_tokens t
       with Not_found -> "<unknown/unexpected token>"
 
-let pp_token: token Pretty.printer = fun ppf ->
+let pp_token: Grammar_tokens.token Pretty.printer = fun ppf ->
   let string s = Pretty.string ppf s
   and print format = Pretty.print ppf format in
-  fun t -> match ~&t with
+  fun t -> match t with
```
nberth, line 98:
```suggestion
  function
```

src/lsp/cobol_parser/text_tokenizer.ml
```diff
@@ -735,7 +738,7 @@ let enable_intrinsics state token tokens =
     let tokens = retokenize_after Enabled_intrinsics state tokens in
     let token, tokens = List.hd tokens, List.tl tokens in
     if show `Tks state then
-      Pretty.error "Tks': %a@." pp_tokens tokens;
+      Pretty.error "Tks': %a@." pp_token'_list tokens;
```
nberth, line 741: Please keep `type tokens = … with_loc list` and `pp_tokens`, that's shorter and perfectly fine for a module internal.

src/lsp/cobol_parser/text_tokenizer.mli
```diff
@@ -62,44 +63,44 @@ val diagnostics
 
 val parsed_tokens
   : Cobol_common.Behaviors.eidetic state
-  -> tokens Lazy.t
```
nberth, line 65: Same here. The `tokens` alias is cleaner.

nberth (changes requested):
I think the only change that is actually needed it the addition or removal of primes for `pp_token` functions. The rest makes things just a little bit too messy.

nberth (approved):
Still not fan of the change to the type names. In the context of the scanner/parser, `type token = … with_loc` was fine. The main issue was only in the name of pretty-printing functions that are exported by the library. But ok anyways.

src/lsp/cobol_parser/keywords/gen_keywords.ml
```diff
@@ -174,6 +174,20 @@ let emit_intrinsic_functions_list ppf =
   Terminal.iter (emit_custom_intrinsics ppf);
   Fmt.pf ppf "@]@\n]@."
 
+let emit_is_intrinsic ppf =
+  if String.equal cmlyname "grammar.cmly" then
```
nberth, line 178: Ouch…!

src/lsp/cobol_parser/keywords/gen_keywords.ml
```diff
@@ -174,6 +174,20 @@ let emit_intrinsic_functions_list ppf =
...
+  let is_intrinsic t =
+    intrinsic (Terminal.attributes t) |> Option.is_some
+  in
+  Fmt.pf ppf "@[<2>let is_intrinsic_token = %s.(function@." tokens_module;
+  Terminal.iter begin fun t ->
+    if is_intrinsic t
+    then Fmt.pf ppf "| %a@." pp_terminal t
+  end;
+  Fmt.pf ppf "| INTRINSIC_FUNC _ -> true@.";
```
nberth, line 187: A bit hardcoded as well. Maybe that case can be taken out, with a refined semantics, renamed like `is_known_intrinsics` (and a need to match `INTRISIC_FUNC` outside of this function).

src/lsp/cobol_parser/grammar_tokens_printer.ml
```diff
@@ -77,7 +77,8 @@ let pp_token: token Pretty.printer = fun ppf ->
     | FIXEDLIT (i, sep, d) -> print "FIXED[%s%c%s]" i sep d
     | FLOATLIT (i, sep, d, e) -> print "FLOAT[%s%c%sE%s]" i sep d e
     | INTERVENING_ c -> print "<%c>" c
-    | tok when Text_keywords.is_intrinsic_token tok ->
+    | INTRINSIC_FUNC _ as tok
+    | tok when Text_keywords.is_known_intrinsic_token tok ->
```
nberth, line 81: Notice that guard applies to both cases (`INTRINSIC_FUNC _ as tok` *and* `tok`). In this case you should just duplicate the for for `printing` (actually using `SPECIALIZED_INTRINSIC_FUNC[%s]` in the second case to distinguish the two kinds of tokens).

nberth (approved):
Thanks!

PR #347, opened 2024-08-07
Add support for dump and listing files

### LINKED ISSUE #328: FR: add support for dump and list files
See https://github.com/gnucobol/vscode_cobol4gnucobol/blob/main/package.json - as the license for the vscode extension here is MIT and the same applies to the referenced repo any definitions can even be copied completely.

This is about recognition, syntax highlighting and outline.

nberth:
Closes #328 & #348

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [0.1.4] Next release
 
 ### Added
-- Support for dump and listing files [#347](https://github.com/OCamlPro/superbol-studio-oss/pull/347)
+- Support for dump and listing files, along with a task attribute for outputting the latter [#347](https://github.com/OCamlPro/superbol-studio-oss/pull/347)
 - Improved information shown on completion [#336](https://github.com/OCamlPro/superbol-studio-oss/pull/336)
 - Configuration flag for caching in storage provided by Visual Studio Code [#167](https://github.com/OCamlPro/superbol-studio-oss/pull/167)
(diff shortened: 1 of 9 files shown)


GitMensch:
Does this include the use of the LSP on those (outline based on this file only)? Otherwise this may only be a partial solve of the referenced issue.

nberth:
> Does this include the use of the LSP on those (outline based on this file only)? Otherwise this may only be a partial solve of the referenced issue.

Not yet, this is only preliminary support; I'll add a separate issue for the outline.

emilienlemaire (approved):

nberth:
> > Does this include the use of the LSP on those (outline based on this file only)? Otherwise this may only be a partial solve of the referenced issue.
> 
> Not yet, this is only preliminary support; I'll add a separate issue for the outline.

Actually that was much more easy to do than I anticipated ;-)

emilienlemaire (approved):

nberth:
@GitMensch Note: this currently hardcodes the format for listing and dump files to an equivalent of `-fformat=VARIABLE`. If such files may include lines longer than 250 chars, then maybe we'd need another source-format with no right margin at all.

GitMensch:
Nice. I'll check once the CI vsix exists, or when there is another binary available.

GitMensch:
Things to check: wide and normal listing -T and -t along with/without symbols and -Xref.

PR #349, opened 2024-08-07
Add support for Codelens requests

### LINKED ISSUE #340: provide optional codelens for references of variables and procedures
If enabled (not sure what the default for a setting like this should be) one would see a codelens "N references" over the definitions of variables and procedures. Clicking on it will then execute `editor.action.findReferences` (show the same popup as the manual way to get there).

Quite similar can be found in the MIT licensed https://github.com/spgennard/vscode_cobol/blob/main/src/vsppcodelens.ts (of course, Copyright still aplies, so if you'd end copying code, then contribution is necessary).

NeoKaios:
Show references count via codelens in data division
Closes #340

### DIFF
--- a/src/lsp/cobol_lsp/lsp_request.ml
+++ b/src/lsp/cobol_lsp/lsp_request.ml
@@ -597,15 +597,33 @@ module PosSet = Set.Make(struct
 
 let codelens_positions ~uri group =
   let filename = Lsp.Uri.to_path uri in
+  let open struct
+    include Cobol_common.Visitor
+    type context =
+      | ProcedureDiv
+      | DataDiv
+      | None
+  end in
+  let set_context context (old, acc) =
+    do_children_and_then (context, acc) (fun (_, acc) -> (old, acc))
+  in
+  let take_when_in context { loc; _ } (current, acc) =
+    if context <> current
+    then skip (current, acc)
+    else
+      let range = Lsp_position.range_of_srcloc_in ~filename loc in
+      skip (context, PosSet.add range.start acc)
+  in
   Cobol_unit.Visitor.fold_unit_group
     object
       inherit [_] Cobol_unit.Visitor.folder
-      method! fold_procedure _ = Cobol_common.Visitor.skip
-      method! fold_qualname' { loc; _ } acc =
-        let range = Lsp_position.range_of_srcloc_in ~filename loc in
-        PosSet.add range.start acc
-        |> Cobol_common.Visitor.do_children
-    end group PosSet.empty
+      method! fold_procedure _ = set_context ProcedureDiv
+      method! fold_paragraph' _ = skip
+      method! fold_data_definitions _ = set_context DataDiv
+      method! fold_procedure_name' = take_when_in ProcedureDiv
+      method! fold_qualname' = take_when_in DataDiv
+    end group (None, PosSet.empty)
+  |> snd
 
 let handle_codelens registry ({ textDocument; _ }: CodeLensParams.t) =
   try_with_main_document_data registry textDocument
--- a/src/lsp/cobol_lsp/lsp_request.ml
+++ b/src/lsp/cobol_lsp/lsp_request.ml
@@ -586,6 +586,55 @@ let handle_document_symbol registry (params: DocumentSymbolParams.t) =
       Some (`DocumentSymbol symbols)
     end
 
+(** { Document Code Lens } *)
+
+module PosSet = Set.Make(struct
+    type t = Position.t
+    let compare (p1: t) (p2: t) =
+      let c = p2.line - p1.line in
+      if c <> 0 then c else p2.character - p1.character
+  end)
+
+let codelens_positions ~uri group =
+  let filename = Lsp.Uri.to_path uri in
+  Cobol_unit.Visitor.fold_unit_group
+    object
+      inherit [_] Cobol_unit.Visitor.folder
+      method! fold_procedure _ = Cobol_common.Visitor.skip
+      method! fold_qualname' { loc; _ } acc =
+        let range = Lsp_position.range_of_srcloc_in ~filename loc in
+        PosSet.add range.start acc
+        |> Cobol_common.Visitor.do_children
+    end group PosSet.empty
+
+let handle_codelens registry ({ textDocument; _ }: CodeLensParams.t) =
+  try_with_main_document_data registry textDocument
+    ~f:begin fun ~doc checked_doc ->
+      let uri = Lsp.Text_document.documentUri doc.textdoc in
+      let rootdir = Lsp_project.(string_of_rootdir @@ rootdir doc.project) in
+      let context = ReferenceContext.create ~includeDeclaration:true in
+      codelens_positions ~uri checked_doc.group
+      |> PosSet.to_seq
+      |> Seq.map begin fun position ->
+        let params =
+          ReferenceParams.create ~context ~position ~textDocument () in
+        let ref_count =
+          lookup_references_in_doc ~rootdir params checked_doc
+          |> Option.fold ~none:0 ~some:List.length in
+        let title = string_of_int ref_count
+                    ^ " reference"
+                    ^ if ref_count > 1 then "s" else "" in
+        let range = Range.create ~end_:position ~start:position in
+        let uri = DocumentUri.yojson_of_t textDocument.uri in
+        let command = Command.create () ~title
+            ~command:"superbol.editor.action.findReferences"
+            ~arguments:[uri; Position.yojson_of_t position] in
+        CodeLens.create ~command ~range ()
+      end
+      |> List.of_seq |> Option.some
+    end
+  |> Option.value ~default:[]
+
 (** {3 Generic handling} *)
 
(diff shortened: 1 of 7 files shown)


src/lsp/cobol_lsp/lsp_request.ml
```diff
@@ -586,6 +586,82 @@ let handle_document_symbol registry (params: DocumentSymbolParams.t) =
       Some (`DocumentSymbol symbols)
     end
 
+(** { Document Code Lens } *)
+
+module PosSet = Set.Make(struct
```
nberth, line 591:
```suggestion
module Positions = Set.Make (struct
```

src/lsp/cobol_lsp/lsp_request.ml
```diff
@@ -586,6 +586,82 @@ let handle_document_symbol registry (params: DocumentSymbolParams.t) =
...
+      method! fold_procedure_name' = take_when_in ProcedureDiv
+      method! fold_qualname' = take_when_in DataDiv
+      method! fold_record_renaming { renaming_name; _ } =
+        take_when_in DataDiv renaming_name
+      method! fold_field_definition' { payload = field; _ } acc =
+        fold_field_definition v { field with field_redefines = None } acc
+        |> skip
+      method! fold_table_definition' { payload = table; _ } acc =
+        fold_table_definition v { table with table_redefines = None } acc
+        |> skip
```
nberth, line 633: Even if that leads to longer code, I think it would be clearer to explicitly fold over selected fields of `field` (resp. `table`). I think "only" `field_qualname`, `field_layout`, `field_conditions`, and `field_redefintions`. Then with an explicit:
```ocaml
ignore (field_redefines, field_leading_ranges,
        field_offset, field_size);
```
that'll help in case we re-define parts of the type.

src/lsp/cobol_typeck/typeck_procedure.ml
```diff
@@ -212,10 +212,19 @@ let references ~(data_definitions: Cobol_unit.Types.data_definitions) procedure
...
-      Visitor.do_children_and_then
-        { acc with current_section = Some s }
-        (fun acc -> { acc with current_section })
+    method! fold_procedure_section ({ section_paragraphs; _ } as s)
+        ({ current_section; _ } as acc) =
+      { acc with current_section = Some s }
+      |> Cobol_common.Visitor.fold_list v section_paragraphs.list
+        ~fold:Cobol_unit.Visitor.fold_procedure_paragraph'
+      |> (fun acc -> { acc with current_section })
+      |> Cobol_common.Visitor.skip
```
nberth, line 221:
```suggestion
      let acc =
        Visitor.fold_list v section_paragraphs.list
          ~fold:Cobol_unit.Visitor.fold_procedure_paragraph'
          { acc with current_section = Some s }
      in
      Visitor.skip { acc with current_section }
```
Please use the reverse style (valid comment in general): `|>` is nice in some places, but doing it the other way round with `@@` in this context helps easily detect what the main behavior is (whether it's a `skip`, or something else). Or like in the suggestion.

src/lsp/cobol_lsp/lsp_request.ml
```diff
@@ -625,13 +625,28 @@ let codelens_positions ~uri group =
...
+                                      field_offset; field_size; field_layout;
+                                      field_conditions; field_redefinitions;
+                                      field_length = _ } acc =
+        ignore(field_redefines, field_leading_ranges, field_offset, field_size);
+        skip @@ begin
+          acc
+          |> Cobol_ptree.Terms_visitor.fold_qualname'_opt v field_qualname
+          |> fold_field_layout v field_layout
+          |> fold_condition_names v field_conditions
+          |> fold_item_redefinitions v field_redefinitions end
```
nberth, line 639:
```suggestion
        skip @@ begin acc
          |> Cobol_ptree.Terms_visitor.fold_qualname'_opt v field_qualname
          |> fold_field_layout v field_layout
          |> fold_condition_names v field_conditions
          |> fold_item_redefinitions v field_redefinitions
        end
```

src/lsp/cobol_lsp/lsp_request.ml
```diff
@@ -625,13 +625,28 @@ let codelens_positions ~uri group =
...
+      method! fold_table_definition { table_field; table_offset; table_size;
+                                      table_range; table_init_values;
+                                      table_redefines; table_redefinitions } acc =
+        ignore(table_offset, table_size, table_init_values, table_redefines);
+        skip @@ begin
+          acc
+          |> fold_field_definition' v table_field
+          |> fold_table_range v table_range
+          |> fold_item_redefinitions v table_redefinitions end
```
nberth, line 648:
```suggestion
        skip @@ begin acc
          |> fold_field_definition' v table_field
          |> fold_table_range v table_range
          |> fold_item_redefinitions v table_redefinitions
        end
```

src/lsp/cobol_typeck/typeck_procedure.ml
```diff
@@ -214,16 +214,16 @@ let references ~(data_definitions: Cobol_unit.Types.data_definitions) procedure
...
+          ~fold:Cobol_unit.Visitor.fold_procedure_paragraph'
+          { acc with current_section = Some s }
+      in
+      Visitor.skip { acc with current_section }
 
     method! fold_procedure_paragraph { paragraph; _ } acc =
-      acc
-      |> Cobol_ptree.Proc_division_visitor.fold_paragraph' v paragraph
-      |> Cobol_common.Visitor.skip
+      Cobol_common.Visitor.skip @@
```
nberth, line 225:
```suggestion
      Visitor.skip @@
```

src/vscode/superbol-vscode-platform/superbol_commands.ml
```diff
@@ -34,6 +34,22 @@ let command id handler =
...
+      | [arg1; arg2] ->
+        let uri = Uri.parse (Ojs.string_of_js arg1) () in
+        let line = Ojs.get_prop_ascii arg2 "line" |> Ojs.int_of_js in
+        let character = Ojs.get_prop_ascii arg2 "character" |> Ojs.int_of_js in
+        let pos = Position.make ~line ~character in
+        let _ = Commands.executeCommand
+            ~command:"editor.action.findReferences"
+            ~args:[Uri.t_to_js uri; Position.t_to_js pos ]
+        in ()
+      | _ -> ()
```
nberth, line 50: I know we don't do that in existing commands, but it may be worth logging some kind of internal warning here. 

Maybe by appending to the log that is associated with the extension (not sure). To do that you'd add `val Superbol_instance.context: t -> Vscode.ExtensionContext.t` to get the context of `instance`, and then get the URI of the log with `Vscode.ExtensionContext.logUri`. There may be functions to append stuff to a file in `Vscode` or in `Node` directly…

Another way would actually be to create an output channel for the extension itself (we have one for the server, but none for "our" client). That can be done like in the OCaml extension with
```ocaml
let extension_oc : Vscode.OutputChannel.t Lazy.t =
  lazy (Vscode.Window.createOutputChannel ~name:"SuperBOL Studio")
```
then with
```ocaml
let lazy oc = extension_oc in
Vscode.OutputChannel.appendLine oc …
```

nberth: @NeoKaios Actually that can be delayed for later if you have something else on your plate right now.

src/vscode/superbol-vscode-platform/superbol_commands.ml
```diff
@@ -35,19 +38,31 @@ let command id handler =
...
+  command command_name @@ Instance
     begin fun _instance ~args ->
       match args with
       | [arg1; arg2] ->
         let uri = Uri.parse (Ojs.string_of_js arg1) () in
         let line = Ojs.get_prop_ascii arg2 "line" |> Ojs.int_of_js in
         let character = Ojs.get_prop_ascii arg2 "character" |> Ojs.int_of_js in
         let pos = Position.make ~line ~character in
+        let lazy oc = extension_oc in
+        let value = "Info: Performing "
```
nberth, line 51: Isn't there a way to use `Printf.sprinf` here?

nberth: Also it's normal behavior, so I don't think we need to log anything here.
The log is just for the warning about misuse of the command.

NeoKaios: Indeed, I only tried `Fmt` and `Pretty`...

nberth: Almost there ;-).

The message here is an indication that something that's internal to the extension went wrong. So the first message where there are two arguments is not needed. And in the other case it's not a usual "Warning", but more a log event that says "Internal warning: unexpected arguments given to <command_name>: expected uri & position, got %s" (with ideally a representation of `arg`; `Obj` may give a `to_string` or similar).

nberth (approved):
Wonderful!

GitMensch:
Agreed - this only misses a README update - but then this may be provided before the (pre)release starting from the Changelog as this is documented there.

PR #350, opened 2024-08-09
Show documentation comment on variable hover

### LINKED ISSUE #339: add documentation (comments) for variables in quick-view/hover
example:

```cobol
      * some line comment
       77 NUM-FLD      PIC S9(35)v9.

       01 FLD-X7.
      *  03 FLD-97     PIC S9(5)v9(32).
         03 FLD-97     PIC S9(02)V9(35).   *> some comment

       PROCEDURE DIVISION.
           MOVE -1234                  TO FLD-97.
           MOVE -1234                  TO NUM-FLD.
```

So when hovering over FLD-97 it would be nice to see "some comment", when hovering over NUM-FLD the "some line comment" would be the appropriate doc.
The line comment gets special for commented code, which  "obviously" should be skipped (which is possibly hard, showing the much less used inline comments would at least be an improvement)

NeoKaios:
Closes #339.

### DIFF
--- a/src/lsp/cobol_lsp/lsp_request.ml
+++ b/src/lsp/cobol_lsp/lsp_request.ml
@@ -454,6 +454,23 @@ let handle_semtoks_full,
 
 (** {3 Hover} *)
 
+let documetation_of_datadef ~(comments: Cobol_preproc.Text.comments) ~filename data_def =
+  let loc = Cobol_data.Item.def_loc data_def in
+  let def_range = Lsp_position.range_of_srcloc_in ~filename loc in
+  let comments = List.rev @@
+    List.filter_map begin fun ({ comment_loc; comment_kind; comment_contents }: Cobol_preproc.Text.comment) ->
+      let com_range = Lsp_position.range_of_lexloc comment_loc in
+      if def_range.start.line = com_range.start.line
+      || def_range.start.line = com_range.start.line + 1
+         && comment_kind ==  `Line
+      then Some comment_contents
+      else None
+    end comments in
+  if List.length comments == 0
+  then ""
+  else Pretty.to_string "\n```cobol\n%a\n```"
+      Fmt.(list ~sep:(any "\n") string) comments
+
 let lookup_data_definition_for_hover cu_name element_at_pos group =
   let { payload = cu; _ } = CUs.find_by_name cu_name group in
   let named_data_defs = cu.unit_data.data_items.named in
@@ -467,7 +484,7 @@ let lookup_data_definition_for_hover cu_name element_at_pos group =
   with Cobol_unit.Qualmap.Ambiguous _ -> raise Not_found
 
 let data_definition_on_hover
-    ?(always_show_hover_text_in_data_div = false)
+    ?(always_show_hover_text_in_data_div = false) ~comments
     ~uri position Cobol_typeck.Outputs.{ group; _ } =
   let filename = Lsp.Uri.to_path uri in
   match Lsp_lookup.element_at_position ~uri position group with
@@ -479,11 +496,12 @@ let data_definition_on_hover
       try
         let data_def, hover_loc
           = lookup_data_definition_for_hover cu_name ele_at_pos group in
+        let doc_comments = documetation_of_datadef ~comments ~filename data_def in
         if always_show_hover_text_in_data_div ||
            not (Lsp_position.is_in_srcloc ~filename position @@
                 Cobol_data.Item.def_loc data_def)
-        then Some (Pretty.to_string
-                     "%a" Lsp_data_info_printer.pp_data_definition data_def,
+        then Some (Pretty.to_string "%a%s"
+                     Lsp_data_info_printer.pp_data_definition data_def doc_comments,
                    hover_loc)
         else None
       with Not_found ->
@@ -536,9 +554,11 @@ let handle_hover ?always_show_hover_text_in_data_div
     registry HoverParams.{ textDocument = doc; position; _ } =
   let filename = Lsp.Uri.to_path doc.uri in
   try_with_main_document_data registry doc
-    ~f:begin fun ~doc:{ artifacts = { pplog; _ }; _ } checked_doc ->
+    ~f:begin fun
+      ~doc:{ artifacts = { pplog; rev_comments = comments; _ }; _ }
+      checked_doc ->
       match data_definition_on_hover ~uri:doc.uri position checked_doc
-              ?always_show_hover_text_in_data_div,
+              ?always_show_hover_text_in_data_div ~comments,
             preproc_info_on_hover ~filename position pplog with
       | None, None ->
           None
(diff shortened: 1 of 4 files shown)


src/lsp/cobol_lsp/lsp_request.ml
```diff
@@ -454,6 +454,23 @@ let handle_semtoks_full,
 
 (** {3 Hover} *)
 
+let documetation_of_datadef ~(comments: Cobol_preproc.Text.comments) ~filename data_def =
```
nberth, line 457: Please break the long lines here.

src/lsp/cobol_lsp/lsp_request.ml
```diff
@@ -454,6 +454,23 @@ let handle_semtoks_full,
...
+  let loc = Cobol_data.Item.def_loc data_def in
+  let def_range = Lsp_position.range_of_srcloc_in ~filename loc in
+  let comments = List.rev @@
```
nberth, line 460:
```suggestion
  let comments =
    List.rev @@
```

src/lsp/cobol_lsp/lsp_request.ml
```diff
@@ -454,6 +454,23 @@ let handle_semtoks_full,
...
+    List.filter_map begin fun ({ comment_loc; comment_kind; comment_contents }: Cobol_preproc.Text.comment) ->
+      let com_range = Lsp_position.range_of_lexloc comment_loc in
+      if def_range.start.line = com_range.start.line
+      || def_range.start.line = com_range.start.line + 1
+         && comment_kind ==  `Line
+      then Some comment_contents
+      else None
+    end comments in
+  if List.length comments == 0
```
nberth, line 469: Please use a `match` to avoid an unnecessary traversal of `List.length` (I doubt it's removed via compiler optimization, even though it could…).

src/lsp/cobol_lsp/lsp_request.ml
nberth: Also substitute `rev_comments` for `comments` (or use a lazy reversal).

src/lsp/cobol_lsp/lsp_request.ml
```diff
@@ -536,9 +554,11 @@ let handle_hover ?always_show_hover_text_in_data_div
     registry HoverParams.{ textDocument = doc; position; _ } =
   let filename = Lsp.Uri.to_path doc.uri in
   try_with_main_document_data registry doc
-    ~f:begin fun ~doc:{ artifacts = { pplog; _ }; _ } checked_doc ->
+    ~f:begin fun
+      ~doc:{ artifacts = { pplog; rev_comments = comments; _ }; _ }
+      checked_doc ->
```
nberth, line 559:
```suggestion
    ~f:begin fun ~doc:{ artifacts = { pplog; rev_comments; _ }; _ } checked_doc ->
```
should do after renaming the `comments` label to `data_definition_on_hover`.

test/lsp/lsp_hover.ml
```diff
@@ -1141,3 +1141,120 @@ let%expect_test "hover-typedef-communication-section" =
...
+    FIELD
+    ```
+    ```cobol
+    PIC X USAGE DISPLAY
+    ```
+    ALPHANUMERIC(1)
+    ```cobol
+    * full line comment
+    *> inline comment
+    ``` |}]
```
nberth, line 1260: To avoid this, you may be able to compare the `pos_fname` field of comment locations against *a* filename obtained from the `def_loc`. (*cf* https://github.com/OCamlPro/superbol-studio-oss/issues/339#issuecomment-2270923754).

I think `Srcloc.forget_preproc ~favor_direction:`Left ~traverse_copies:true ~traverse_replaces:false def_loc` (whose short equivalent is `Srcloc.as_lexloc def_loc` should do it.

src/lsp/cobol_lsp/lsp_request.ml
nberth: Also I don't think the case of multiple line comments is in the tests. Also what about an item with both a line and an inline comment (I guess the latter should prevail).

src/lsp/cobol_lsp/lsp_request.ml
```diff
@@ -454,22 +454,29 @@ let handle_semtoks_full,
...
+      List.rev @@
+      List.filter_map
+        begin fun { comment_loc; comment_kind; comment_contents } ->
+          let com_range = Lsp_position.range_of_lexloc comment_loc in
+          if def_range.start.line = com_range.start.line
+          || def_range.start.line = com_range.start.line + 1
+             && comment_kind ==  `Line
+          then Some comment_contents
+          else None
+        end rev_comments in
```
nberth, line 475:
```suggestion
      List.filter_map begin fun { comment_loc; comment_kind; comment_contents } ->
          let com_range = Lsp_position.range_of_lexloc comment_loc in
          if def_range.start.line = com_range.start.line
          || def_range.start.line = com_range.start.line + 1
             && comment_kind ==  `Line
          then Some comment_contents
          else None
      end rev_comments
    in
```

nberth (approved):
Thanks!

test/lsp/lsp_hover.ml
```diff
@@ -1141,3 +1141,116 @@ let%expect_test "hover-typedef-communication-section" =
...
+      13            STOP RUN.
+      14
+    ```cobol
+    STRUCT
+    ```
+    Group of 2 subfields
+    Size: 16 bits
+    ```cobol
+    * full line comment
+    *> inline comment
```
nberth, line 1180: I don't think the two kinds of comments should be shown. In this case I think only the inline comment should be considered. Plus not inside a COBOL code block either, and without the comment markers (juste assuming it's plain text — that could be checked later via some heuristics).

src/lsp/cobol_lsp/lsp_request.ml
```diff
@@ -458,25 +458,25 @@ let documetation_of_datadef ~rev_comments ~filename data_def =
...
+        then (fst acc, Some comment_contents)
+        else acc
+      end (None, None) rev_comments
     in
-    match comments with
-    | [] -> ""
-    | _ -> Pretty.to_string "\n```cobol\n%a\n```"
-             Fmt.(list ~sep:(any "\n") string) comments
+    match inline, full_line with
+    | Some comment, _ -> "\n---\n" ^ String.sub comment 3 (String.length comment - 3)
```
nberth, line 477: I think `*>Something` should render `Something`, not `omething` ;-)
Also: I think several line comments should be concatenated. I think the current behavior is to take only the last line.

Makes me think we should remove markers from `comment_contents` (I'll post an issue about that).

src/lsp/cobol_lsp/lsp_request.ml
```diff
@@ -454,6 +454,30 @@ let handle_semtoks_full,
 
 (** {3 Hover} *)
 
+let documetation_of_datadef ~rev_comments ~filename data_def =
```
nberth, line 457:
```suggestion
let documentation_of_datadef ~rev_comments ~filename data_def =
```
(or else just `doc_at_datadef`).

PR #367, opened 2024-09-03
more explanations on settings

GitMensch:
and back-reference for "show copybooks on hover"

this originates on me wondering why it did not work "as suggested by the README".

Question: Wouldn't it be reasonable for the LSP to have a list of the file names and if there's only a single one to match the copybook for hover/show definition`?
This would lead to most environments working "out of the box".

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,8 +83,9 @@ From here, you can notably configure:
   To configure this setting, you will need to select `Edit in settings.json`. This is
   a list, where each entry describes a file extension if the copybook name cannot
   be found as-is (for example `COPY "mycpy.lib"`).
-  In SuperBOL, the `default` option does not correspond to GnuCOBOL's default, but to "cpy" and
-  "cbx" only.
+  In SuperBOL, the `default` option corresponds to GnuCOBOL's default, which
+  is `"cpy"`, `"cbl"`, and `"cob"` (searched in upper-case first, and then in
+  lower-case).
 
 ## Syntax diagnostics
 
--- a/README.md
+++ b/README.md
@@ -51,26 +51,26 @@ be presented with a screen that resembles:
 
 From here, you can notably configure:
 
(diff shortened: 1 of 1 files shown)


GitMensch:
requesting review from @nberth

README.md
```diff
@@ -79,6 +79,13 @@ From here, you can notably configure:
 
   ![Editing copybook paths in `.vscode/settings.json`](./assets/superbol-editing-copybooks-path-in-vscode-settings.png)
 
+- The copybook extensions `"superbol.cobol.copyexts"`.  
+  To configure this setting, you will need to select `Edit in settings.json`. This is
+  a list, where each entry describes a file extension if the copybook name cannot
+  be found as-is (for example `COPY "mycpy.lib"`).
+  In SuperBOL, the `default` option does not correspond to GnuCOBOL's default, but to "cpy" and
+  "cbx" only.
```
nberth, line 87: Is GnuCOBOL's default list actually `cpy cbl cob` as hinted [here](https://github.com/OCamlPro/gnucobol/blob/fe28973b030f13b34c52d36df6b928069de21f98/cobc/cobc.c#L9077)?

GitMensch: yes, first in upper, then in lower-case

GitMensch:
@nberth: any updates on this? I'm not sure if  #368 is likely to be finished soon and have a "frontend" in the vscode extension (it definitely sounds interesting), but with all the other important changes since the last release I'd like to see 0.1.4 going out "rather soon" - and thise one seems like a low hanging fruit before...

README.md
nberth, line 87:
```suggestion
  In SuperBOL, the `default` option corresponds to GnuCOBOL's default, which
  is `"cpy"`, `"cbl"`, and `"cob"` (searched in upper-case first, and then in
  lower-case).
```

PR #368, opened 2024-09-09
Support for COBOL CFG display with `graphviz` and `d3`

NeoKaios:
- Adds `ocaml-graph` dependency to generated a dot graph of the CFG of a COBOL program / section
- Adds 2 commands to acces a dot svg graph and a arc-reprensentation of the CFG via Vscode webviews
- Adds context menus to access those graphs


src/lsp/cobol_cfg/cfg_builder.ml
```diff
@@ -0,0 +1,486 @@
...
+(******************************************************************************)
+
+open Cobol_unit
+open Cobol_common.Srcloc.INFIX
+open Cobol_common.Srcloc.TYPES
+open Cobol_unit.Types
+open Cobol_common.Visitor
+module NEL = Cobol_common.Basics.NEL
+
+module Options = struct
```
nberth, line 18: I'd rather not depend on `yojson` here if possible. This `type option` should just be exposed by `Cobol_cfg` and constructed by "users" of that library. This means at the moment parsing the JSON should only be done in `Cobol_lsp` I think.

src/lsp/cobol_lsp/lsp_request.ml
```diff
@@ -825,6 +872,12 @@ let on_request
     | UnknownRequest { meth = "superbol/getProjectConfiguration";
                        params = Some param } ->
         handle_get_project_config_command param registry
+    | UnknownRequest { meth = "superbol/CFG";
```
nberth, line 875: Better named `superbol/getCFG` I think.

src/lsp/superbol_free_lib/vscode_extension.ml
```diff
@@ -583,13 +583,26 @@ let contributes =
         ~command:"superbol.coverage.reload"
         ~title:"Update Coverage"
         ~category:"SuperBOL";
+      Manifest.command ()
+        ~command:"superbol.cfg.open"
+        ~title:"Open a CFG of the current program"
+        ~category:"SuperBOL";
+      Manifest.command ()
+        ~command:"superbol.cfg.open.arc"
+        ~title:"Open an arc CFG of the current program"
```
nberth, line 592: "Arc CFG"? Instead I think "Show control-flow as an arc-diagram" may fit better: shorter, straight to the point. (Same above with the prefix of this sentence).

src/vscode/superbol-vscode-platform/superbol_cfg_explorer.ml
```diff
@@ -0,0 +1,284 @@
...
+(*  LICENSE.md file in the root directory of this source tree.            *)
+(*                                                                        *)
+(*                                                                        *)
+(**************************************************************************)
+
+open Vscode
+
+let read_whole_file filename =
+  (* open_in_bin works correctly on Unix and Windows *)
+  let ch = open_in_bin filename in
```
nberth, line 19: I think there might be some exception to handle here (maybe re-raise, but `ch` might need to be closed anyways in this case).

src/vscode/superbol-vscode-platform/superbol_cfg_explorer.ml
```diff
@@ -0,0 +1,284 @@
...
+let update_graph ~uri graph =
+  let filename = Uri.path uri in
+  match Hashtbl.find_opt webview_panels filename with
+  | Some (wvp, _) ->
+    Hashtbl.replace webview_panels filename (wvp, graph)
+  | None -> ()
+
+(* CLICK ON NODE *)
+
+let ignore_next_selection_change = ref false
```
nberth, line 111: What is this state attached to? a window? an editor? a webview?

src/lsp/cobol_cfg/cobol_cfg.ml
```diff
@@ -0,0 +1,23 @@
...
+(* of this source tree.                                                   *)
+(*                                                                        *)
+(**************************************************************************)
+
+module INTERNAL = struct
+  module Types = struct
+  end
+end
+
+module Builder = struct
```
nberth, line 19: I think it will make more sense to add a stripped-down `cfg_builder.mli` and then only have `module Builder = Cfg_builder`.

src/vscode/superbol-vscode-platform/superbol_cfg_explorer.mli
```diff
@@ -0,0 +1,19 @@
...
+(*                                                                        *)
+(*  All rights reserved.                                                  *)
+(*  This source code is licensed under the MIT license found in the       *)
+(*  LICENSE.md file in the root directory of this source tree.            *)
+(*                                                                        *)
+(*                                                                        *)
+(**************************************************************************)
+
+val open_cfg
+  : ?typ: [< `Arc | `Dot > `Dot]
```
nberth, line 16: For clarity of this interface I suggest to simply declare `type cfg_type = Graphviz | D3_arc_diagram` (adjust those constructors if they don't correspond).

src/vscode/superbol-vscode-platform/superbol_cfg_explorer.ml
NeoKaios: This has been reworked

src/lsp/cobol_cfg/package.toml
```diff
@@ -0,0 +1,85 @@
...
+#   base-unix = { libname = "unix", version = ">=base" }
+[dependencies]
+cobol_common = "version"
+cobol_config = "version"
+cobol_data = "version"
+cobol_indent = "version"
+cobol_parser = "version"
+cobol_typeck = "version"
+superbol_preprocs = "version"
+superbol_project = "version"
```
nberth, line 63: Does `cobol_cfg` really depend on that many packages. In any case it should not depend on any `superbol_*`.

src/lsp/cobol_cfg/cfg_options.ml
```diff
@@ -0,0 +1,43 @@
...
+  graph_name: string option;
+  hide_unreachable: bool;
+  collapse_fallthru: bool;
+  shatter_hubs: int option;
+  transformation: transformation option;
+  hidden_nodes: int list;
+  split_nodes: int list;
+}
+
+let create
```
nberth, line 25: We don't use this kind of creation function in these libraries. Directly creating the record should be straightforward enough.

src/lsp/cobol_cfg/cobol_cfg.ml
```diff
@@ -0,0 +1,15 @@
+(**************************************************************************)
+(*                                                                        *)
+(*                        SuperBOL OSS Studio                             *)
+(*                                                                        *)
+(*  Copyright (c) 2022-2023 OCamlPro SAS                                  *)
```
nberth, line 5: Something is wrong about this header. For some time I think we added the `2024`, didn't we?

nberth: Ah! this part of the code is not AGPL. Please use the same header as in `cfg_jumps.ml` for every file in this library.

src/lsp/cobol_cfg/cfg_jumps.ml
```diff
@@ -0,0 +1,161 @@
...
+module Jumps = Set.Make(struct
+    type t = jumps
+    let compare j1 j2 =
+      let to_int = function
+        | Go _ -> 0
+        | GoDepending _ -> 1
+        | Perform _ -> 2
+        | Call _ -> 3
+        | Entry _ -> 4
+      in match j1, j2 with
```
nberth, line 34:
```suggestion
      in
      match j1, j2 with
```

src/lsp/cobol_cfg/cfg_jumps.ml
```diff
@@ -0,0 +1,161 @@
...
+      | Go qn1, Go qn2
+      | GoDepending qn1, GoDepending qn2
+      | Perform qn1, Perform qn2 -> Cobol_ptree.compare_qualname qn1 qn2
+      | Call cp1, Call cp2 -> Cobol_ptree.compare_call_prefix cp1 cp2
+      | Entry en1 , Entry en2 -> Cobol_ptree.compare_entry_stmt en1 en2
+      | _ -> to_int j2 - to_int j1
```
nberth, line 40: Why not use ppx-derivers for those? (as we used in other parts of the code)?

src/lsp/cobol_cfg/cfg_jumps.ml
```diff
@@ -0,0 +1,161 @@
...
+(*     OCAMLPRO-NON-COMMERCIAL license.                                       *)
+(*                                                                            *)
+(******************************************************************************)
+
+open Cobol_unit
+open Cobol_unit.Types
+open Cobol_common.Visitor
+open Cobol_common.Srcloc.INFIX
+
+type qualname = Cobol_ptree.qualname
```
nberth, line 16: No need for this alias.

src/lsp/cobol_cfg/cfg_builder.mli
```diff
@@ -0,0 +1,21 @@
...
+(*     OCAMLPRO-NON-COMMERCIAL license.                                       *)
+(*                                                                            *)
+(******************************************************************************)
+
+type graph = {
+  name: string;
+  string_repr_dot: string;
+  string_repr_d3: string;
+  nodes_pos: (int * Cobol_ptree.srcloc) list
+}
```
nberth, line 16: For later: this `graph` should actually be the representation as in `ocamlgraph`. The export to graphviz & d3 should be part of the `cobol_lsp` library itself. The `srcloc`s can probably be attached as node label/attributes in the `ocamlgraph` representation, and then extracted into this list in the code that build the reply in `cobol_lsp`.

src/lsp/cobol_cfg/cfg_builder.mli
```diff
@@ -0,0 +1,21 @@
...
+
+val make
+  : options:Cfg_options.t
+  -> Cobol_typeck.Outputs.t
+  -> graph list
```
nberth, line 21: Can you describe a bit more what this list is made of.?

src/lsp/cobol_cfg/cfg_builder.ml
```diff
@@ -0,0 +1,542 @@
...
+      "qn with more than 2 qualification levels cannot \
+       come from a paragraph or section"
+
+let qn_to_fullname qn =
+  let name, qual = qn_to_strings qn in
+  if qual == ""
+  then name
+  else name ^ " IN " ^ qual
+
+let incr ref =
```
nberth, line 39: This redefinition is to be avoided. Maybe this could be named `next_int` or something like this.

Could be even better to use [`Seq.ints 0`](https://ocaml.org/manual/5.2/api/Seq.html#1_Sequencesofintegers)

src/lsp/cobol_cfg/cfg_builder.ml
nberth, line 1: I only skimmed through this file, which appears fine as a draft. Some changes will be needed (possibly later) to avoid doing the actual export in this library (which should mostly be about *internal* CFG representation and construction, with possibly a `Cobol_cfg.As_dot` module which may be helpful in other contexts).

assets/cfg-d3-renderer.html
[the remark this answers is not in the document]
NeoKaios: Which one ? OCAMLPRO-NON-COMMERCIAL license ?

src/lsp/cobol_cfg/cfg_jumps.ml
NeoKaios: Because I don't know how to use them :upside_down_face:

src/lsp/cobol_lsp/lsp_request.ml
```diff
@@ -133,6 +133,89 @@ let handle_get_project_config_command param registry =
...
+          end nodes_pos in
+        `Assoc [
+          ("string_repr_d3", `String string_repr_d3);
+          ("string_repr_dot", `String string_repr_dot);
+          ("nodes_pos", `Assoc nodes_pos);
+          ("name", `String name);]
+      in
+      Some (`List (List.map yojsonify graphs))
+    end
+  |> Option.value ~default:(`List [])
```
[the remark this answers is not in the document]
NeoKaios: raising here makes the client fails, so promise on the VSCode side rejects which is not handled atm
To fail gracefully, I added thoses falsy values to detect error cases on the vscode side.
I could try handling promise rejection to get rid of those default

assets/cfg-d3-renderer.html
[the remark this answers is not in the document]
nberth: I think headers with the contents below should do:
```js
// -----------------------------------------------------------------------
//
//                        SuperBOL OSS Studio
//
//
//  Copyright (c) 2024 OCamlPro SAS
//
//  All rights reserved.
//  This source code is licensed under the MIT license found in the
//  LICENSE.md file in the root directory of this source tree.
//
// -----------------------------------------------------------------------
```

src/lsp/cobol_cfg/cfg_builder.mli
```diff
@@ -0,0 +1,149 @@
...
+  | Go
+
+module Edge: sig
+  type t = edge
+
+  val compare : 'a -> 'a -> int
+  val default : edge
+end
+
+module Cfg :
```
nberth, line 50: Here you probably just need to re-use the `Graph.Sig.P` module type. That would be something like
```ocaml
module Cfg: Graph.Sig.P
  with type V.t = Node.t
   and type V.label = Node.t
   and type E.t = Node.t * Edge.t * Node.t
   and type E.label = Edge.t
```

nberth: The declarations (modules `Node`, `Edge` and `Cfg`, along with types `node_type`, `node` and `edge` should be in a separate module (I suggest `Cfg_types`, bound as `Cobol_cfg.Types`).

src/lsp/cobol_cfg/cfg_options.ml
```diff
@@ -0,0 +1,23 @@
...
+(******************************************************************************)
+
+type transformation =
+  | Descendents of int
+  | Neighborhood of int
+
+type t = {
+  hide_unreachable: bool;
+  collapse_fallthru: bool;
+  shatter_hubs: int option;
```
nberth, line 18: `shatter_hubs` is a kind of threshold on a measure on produced graphs. So its name should be correspond do that. Is this something like `node_degree_upper_limit` (counting every arc entering/leaving a node)?

src/lsp/cobol_cfg/cfg_options.ml
```diff
@@ -0,0 +1,23 @@
...
+  transformation: transformation option;
+  hidden_nodes: int list;
+  split_nodes: int list;
```
nberth, line 21: Note sure what those two last lists correspond to. What do the `int`s correspond to?

NeoKaios: They correspond to node ids

nberth: Could you add a comment to specify where these IDs come from? Do they come from a previous CFG construction? (there's an integrity issue here, but that may still be ok for now).

src/lsp/cobol_lsp/lsp_request.ml
```diff
@@ -133,6 +133,55 @@ let handle_get_project_config_command param registry =
...
+      try to_assoc @@ member "render_options" params with Type_error _ -> [])
+  in
+  let textDoc = TextDocumentIdentifier.create ~uri:(DocumentUri.of_path uri) in
+  try_with_main_document_data registry textDoc
+    ~f:begin fun ~doc:_ checked_doc ->
+      let jsoono =
+        Lsp_cfg.doc_to_cfg_jsoono ~filename:uri ~name ~options checked_doc
+      in Some jsoono
+    end |>
+  Option.get
```
nberth, line 150: Not sure why there's a need for `Option.get` here. Like most other requests the payload of replies should also be made optional to account for improper uses of the requests, and those cases dealt with on the client side.

(That'll be fixed later though)

nberth:
Thanks!

PR #370, opened 2024-10-14
Add SQL preprocessor

NeoKaios:
An up-to-date with master version of #307
Also contains various bugfixes found while launching Gixsql testsuite

Issue #327, opened 2024-07-25
bind variables should be recognized (goto/show definition)


GitMensch:
bind variables (= variables that are from the COBOL part) are always prefixed with a double-colon:

```sql
EXEC SQL
    DELETE FROM TAB WHERE TAB-ID = :VAR
END-EXEC
```

`VAR` should be referenced as COBOL variable and therefore goto/show definition should apply to it.

GitMensch:
Is that hard enough to postpone to ß - or is the later "in sight" already?

GitMensch:
The bin variables are normally only in `WORKING STORAGE` or `LINKAGE` within `EXEC SQL DECLARE` or in copybooks included by `EXEC SQL INCLUDE` - and always "defined above" (no forward declarations).

So yes, in general the lookup seems "dumb" (and minimal more complex for the `EXEC SQL INCLUDE copybookname` case, but that's either understood by the LSP already or not [in this case it would be good to add it in general - that's just a COPY without REPLACE option done by the sql preprocessor before the compiler sees it]).

GitMensch:
Friendly ping as this is one of the "more missing" editing features compared to a setup with the cobol extension from spgennard.

PR #375, opened 2024-10-23
Detect COBOL variable in `EXEC SQL`, provides go to def, show refs and hover

NeoKaios:
Add-on to #370,
reactivate Sql preproc in the LSP and add references and location to cobol vars in sql exec blocks

The completion seems to fails inside such exec blocks -> the parser needs to be more robust
The new completion test needs to pass before merging this


GitMensch:
This _seems_ to target #327, no? If yes I suggest to reference that in the commit message and/or in this PR as related issue.

nberth:
This PR introduces dependencies between libraries that I'd like to avoid as they would impair further extensibility of the LSP server. In addition the parser for SQL is not hardened against errors in SQL syntax. Those issues need to be fixed before this PR can be merged.

GitMensch:
I still wonder if it wouldn't be "simple" to not fully parse the SQL stuff (which can be quite useful if it provides completions and co) but just scan it for any doublecolon+word - and add those as a reference for "word"....

nberth:
> Should it?

Implicitly declared variables are not supported yet: only references to COBOL variables for which there is a corresponding COBOL declaration can be navigated (and renamed) in this way. I guess this will be future work (and this limitation will need to be documented in the `README.md` — that does not even mention SQL yet).

nberth:
> for the cursors, sure - but FL1 and FLD2 are indeed variables that are part of `WORKING-STORAGE`, even separate fields
> 
> ```cobol
>            EXEC SQL BEGIN DECLARE SECTION END-EXEC.
>       *
>        01 DB                 PIC  X(010).
>        01 FLD1               PIC  X(018).
>        01 FLD2               PIC  X(026).
>       * 
>        01 MYTAB.
>          03 MYKEY.
>            05 FLD3           PIC  X(006).
>            05 FLD4           PIC  X(008).
>       *
>          03 MYDATA.
>             05 ...
>       * 
>            EXEC SQL VAR FLD3            IS CHAR      END-EXEC.
>            EXEC SQL VAR FLD4            IS CHAR      END-EXEC.
>       * 
>            EXEC SQL END DECLARE SECTION END-EXEC.
>            EXEC SQL INCLUDE SQLCA END-EXEC.
>       *
>            EXEC SQL AT :DBS
>               DECLARE MYCUR CURSOR FOR
>                  SELECT
>                         FLD1, FLD2 FROM TAB
>                  WHERE
>                           TFL3
>                        || TFL4
>                  >=
>                           :FLD3
>                        || :FLD4
>                  ORDER BY
>                           TFL3
>                        || TFL4
>                  ASC
>            END-EXEC.
> ```

All of this snippet is in `WORKING-STORAGE` section, right?

GitMensch:
yes (that's the place Pro*COB expects it; it possibly doesn't care and only searches for EXEC SQL ... END-EXEC and EXEC SQL DECLARE ...  END DECLARE)

Issue #365, opened 2024-09-03
provide symbols in workspace


GitMensch:
With the current released vsix the "symbols in editor" (starting with `@` in the omnibar) works as expected.
But the "symbols in workspace" (starting with `#` in the omnibar) shows that there are no symbols provided for that yet.

I suggest to include the following:
* `PROGRAM-ID`s; `FUNCTION-ID`s; `METHOD-ID`s
* `EXTERNAL` variable-names, maybe only level 01
* `EXTERNAL` file-names
* from copybooks - possibly later:
    * data-names with 01/77/78 level numbers
    * `SECTION` names

GitMensch:
> No this is not supported yet. This apparently requires some ability to precisely scan a whole workspace, which we anticipate to be a rather costly operation, and therefore needs to be designed with care on the LSP server's side to avoid too many noticeable impact on performance.

_Originally posted by @nberth in https://github.com/OCamlPro/superbol-studio-oss/pull/377#discussion_r1815254083_

The way to do that (best practice when spgennard added something similar) is to create a command for scanning the whole workspace. This can be executed manually. There should be a configuration to automatically run this once at startup.

The result would be put somewhere (similar as ctags would do it), possibly even one file per workspace file (the file date will be included so that if the file did not changed that db entry is not updated and the LSP does not need to scan it).

If a user opens a file (= the old results from the db can be used directly and then a scan for this background be done - so the general performance for opening big files would also increase) this database is always updated (so even without doing the global scan once the db will increase step by step - as files are opened).

The place to write this database should default to the workspace/extension storage, but be able to be configured for a different place **which then also needs to be prepared for multi-user access** [huge workspace with one shared db].

Issue #354, opened 2024-08-20
provide amount of references in the variable/procedure hover


GitMensch:
While #340 is not yet released and the README wasn't update to include it, I expect to see the number of references as codeLens in the definition - at least if this definition/use is in the same file.

If there's a variable (or procedure) defined in a copybook, this will likely be not available "directly".

I suggest to add the number of references also in the hover over a variable/procedure - this allows a programmer to quickly see which effect the use of that may have. If that is possible, then it would be nice for variables to see also the kind of references, something like "references: 12" for procedures and "references: 4 target, 3 source".

I _guess_ that the LSP has that information (otherwise `cobc -t $tmp/prog.lst -Xref` would have that.

GitMensch:
Hm, isn't this implemented _and_ shipped already?

GitMensch:
... but: the LSP (@lefessan ?) seems to miss a bunch of references, see this sample program:

```cobol

        IDENTIFICATION DIVISION.
        program-id. rw-test.
        ENVIRONMENT DIVISION.
        INPUT-OUTPUT SECTION.
        FILE-CONTROL.  *> only needed if not compiling with -std=mf-strict
           select report-file assign "rpt.txt"
               organization line sequential.
        data division.
        file section.

        fd report-file
           report is report-1.
       working-storage section.
       01 n comp pic 99 value 0.
       01 test-table.
         02 test-data.
           03 pic 9999 value 1001.
           03 pic 9999 value 1002.
           03 pic 9999 value 1003.
           03 pic 9999 value 2004.
           03 pic 9999 value 2005.
           03 pic 9999 value 2006.
         02 test-entry redefines test-data pic 9999 occurs 6.
       01 report-entry.
           03 test-group pic 9.
           03 test-value pic 999.
       report section.
       rd report-1
        control is test-group
        page limit 50 lines.
       01 rw-detail type de.
         02 line plus 1.
           03 grp column 1 pic 9 source test-group.
           03 val column 4 pic zz9 source test-value.
       01 rw-foot type cf test-group.
         02 line plus 1.
           03 column 1 pic x(6) value "-  ---".
         02 line plus 1.
           03 column 1 pic 9 source test-group.
           03 s-mm column 4 pic zz9        *> s.mm defined
               sum test-value
               reset test-group.
         02 line plus 1 pic x value space.
       procedure division.
       declaratives.
       decl-rpt section.
           use before reporting rw-foot.
[...]

PR #377, opened 2024-10-24
More documentation

nberth:


README.md
```diff
@@ -146,6 +147,13 @@ select  `Go to Definition` (or press <kbd>F12</kbd>).
 
 ![Go to Definition](./assets/superbol-goto-definition.gif)
 
+> [!NOTE]
+> (Temporary limitation)
+>
+> At the moment, definitions that belong to *some* sections of the
+> data division, and some references to data items, are ignored by the
+> extension.  They will be covered by the first stable release.
```
emilienlemaire, line 155:
```suggestion
> extension. They will be covered by the first stable release.
```

engboris, line 155:
```suggestion
> extension. They will be covered by the first stable release.
```

README.md
```diff
@@ -191,6 +206,50 @@ replacement by a `REPLACE` directive in the same way.
...
+Sections and paragraphs of the procedure division can also be renamed
+in the same way.  
+The same limitations as for [Go to Definition](#go-to-definition)
+apply to this feature.
+
+## Exploring the Control-flow
+
+Navigating a graphical representation of a COBOL program's
+control-flow proves invaluable when in comes to decypher its overall
+logic.  To do this, open the command palette (or type
```
emilienlemaire, line 231:
```suggestion
logic. To do this, open the command palette (or type
```

README.md
```diff
@@ -191,6 +206,50 @@ replacement by a `REPLACE` directive in the same way.
...
+<kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>P</kbd>), and select `SuperBOL:
+Show control-flow` (you can also right click and select `Show
+control-flow` in the menu).  You are then presented with a list of
```
emilienlemaire, line 234:
```suggestion
control-flow` in the menu). You are then presented with a list of
```

README.md
```diff
@@ -191,6 +206,50 @@ replacement by a `REPLACE` directive in the same way.
...
+
+![CFG Explorer](./assets/superbol-cfg-explorer.gif)
+
+Various settings are provided to tune the rendering of CFGs.
+
+![CFG Explorer Collapse Fallthrough](./assets/superbol-cfg-explorer-collapse-fallthrough.gif)
+
+### CFG as an arc diagram
+
+A rendering of CFGs as arc diagrams is also available.  In this
```
emilienlemaire, line 247:
```suggestion
A rendering of CFGs as arc diagrams is also available. In this
```

emilienlemaire (approved):
Looks good to me, just a few too many spaces :smiley:

README.md
```diff
@@ -191,6 +206,50 @@ replacement by a `REPLACE` directive in the same way.
...
+representation, sections and paragraph named are layed out vertically,
```
engboris, line 248:
```suggestion
representation, sections and paragraphs named are laid out vertically,
```

engboris (approved):
Some typos

nberth:
> Looks good to me, just a few too many spaces 😃

That's a habit I kept from the time I edited LaTeX: it's actually very useful to distinguish sentences, and has no impact on markdown semantics.

README.md
```diff
@@ -116,7 +116,8 @@ selected suggestion.
 > (Temporary limitation)
 >
 > Suggestions of user-defined words may not comprise symbols that are
-> defined in some sections of the data division.
+> defined in some sections of the data division.  They do not include
```
GitMensch, line 119: I really do wonder what "some sections" means... also it would be reasonable to say here something like "user-defined words from configured copybooks [link] are included" (if this is the case, what I guess).

README.md
```diff
@@ -146,6 +147,13 @@ select  `Go to Definition` (or press <kbd>F12</kbd>).
 
 ![Go to Definition](./assets/superbol-goto-definition.gif)
```
GitMensch, line 148: One of the navigation features of vscode is "go to symbol in workspace" - is this supported and if not hard to add (#365)?

README.md
```diff
@@ -191,6 +213,50 @@ replacement by a `REPLACE` directive in the same way.
...
+
+A rendering of CFGs as arc diagrams is also available.  In this
+representation, named sections and paragraphs are laid out vertically,
+and arcs show the direction of control-flow between them.
+
+![CFG Explorer as Arc Diagram](./assets/superbol-cfg-explorer-arc.gif)
+
 ## Debugging
 
 In order to debug a COBOL program, you first need to run a *build
```
GitMensch, line 262: This section misses things like attaching, setting program/procedure breakpoints, watchpoints, ...

If you change the extension to keep the debugging in its own separate extension (with a dependency in this extension to the debug extension but not vice-versa if the other extension does not use the LSP), then you'd only need a teaser here and can reference to that's extension's README, which should already include parts of this.

Also I'd highly suggest to not only hint to the debug build task, because that is only useful for small/testing programs [
but there it is _very_ useful, especially for beginners]. With bigger environments you'll have either external build flows or (most preferably) done it by `make`.

For this scenario it is important to document which options need to be passed to cobc (I _think_ for GC 3.2+ the `-g` option is enough).


As the coverage handling is also part of that extension, the same applies there - but in any case I'd suggest to have at least a screenshot of that here.

README.md
nberth: No this is not supported yet. This apparently requires some ability to precisely scan a whole workspace, which we anticipate to be a rather costly operation, and therefore needs to be designed with care on the LSP server's side to avoid too many noticeable impact on performance.

README.md
```diff
@@ -169,9 +177,23 @@ every reference to this item.
...
+> In addition to limitations mentioned in [Go to
+> Definition](#go-to-definition), references that occur in
+> `EXEC`/`END-EXEC` blocks are also not taken into account yet.
+
+### Reference Information
+
+The extension shows inline reference information above definitons of
+data items and elements of the procedure division.  
+The same limitations as for [Go to Definition](#go-to-definition)
+apply.
```
GitMensch, line 189: please add a sscreenshot showing the feature here as well; I _guess_ that possibly means #354 can be closed

nberth: Thanks indeed the screenshot is missing here.
#354 would probably be easy to implement (including the reference count in the generated information shown on hover), but it's not done yet.

README.md
nberth: I've added basic details about the additional flags passed to `cobc`.

Regarding dependencies, at the moment the debug extension is tuned with a `preLaunchTask` that directs to a build task of the main extension; this is mainly to help newcomers (they can just press <kbd>F5</kbd> to automatically compile and run programs). I'd like to keep the extensions separate, but for now we found no convincing way to do that and keep the aforementioned behavior (it's also still debatable whether this fully automated behavior is really worth the hassle — see also the other debug-related issues).

At the moment I'm trying to do a quick release (long overdue); so aspects mentioned in this comment will be addressed later.

Note we don't rely on the debug extension to show coverage information (there's no reason to have those two features so tied together I think): a dedicated gcov-viewer extension that we found simpler/more maintainable than the coverage-related code of the debug extension is used instead.

GitMensch: > a dedicated gcov-viewer extension that we found simpler/more maintainable than the coverage-related code of the debug extension is used instead.

That's good - can you keep that in a separate (depending = automatically installed [as long as you're connected to the internet]) extension or does that needs to "included" as well?

Please create an issue to drop the coverage stuff from the debugging extension (because it is not needed any more and is "too much" there as well).

PR #413, opened 2025-05-23
Make SuperBOL configs CamelCase in VSCode extension

### LINKED ISSUE #407: configuration: consider moving from partially hyphen-separated to all camel-case
Currently some settings use camel-case (like `cacheInGlobalStoratge`) while others use hyphens (like `display-variable-attributes` or `lsp-path`).

In general it is considered "good", to choose _one_ way of doing it and stick with it.

The defacto-standard for vscode is camel-case. This is also recognizable in the settings ui as this splits the words on "Camel", while it doesn't do that for hyphens.

And of course: the move from "pre-release" to "release", which happens with the next version is a perfect time for renaming settings - no time better than the current :-)

engboris:
Closes #407

### DIFF
--- a/src/lsp/superbol_free_lib/vscode_extension.ml
+++ b/src/lsp/superbol_free_lib/vscode_extension.ml
@@ -313,7 +313,8 @@ let contributes =
            (* Note: entries that belong to the `superbol.cobol` section is
               transmitted to the LSP server, and needs to exactly mirror the
               `[cobol]` section in `superbol.toml`.  There, convention for keys
-              is to use dash-separated words in lowercase.  *)
+              is to use dash-separated words in lowercase in order to keep
+              consistency with GnuCOBOL's configurations.  *)
 
            Manifest.PROPERTY.enum "superbol.cobol.dialect"
              ~cases:Cobol_config.DIALECT.all_canonical_names
@@ -324,7 +325,7 @@ let contributes =
              ~default:(`String Cobol_config.(DIALECT.to_string Default))
              ~order:1;
 
-           Manifest.PROPERTY.enum "superbol.cobol.source-format"
+           Manifest.PROPERTY.enum "superbol.cobol.sourceFormat"
              ~title:"Source Reference-format"
              ~markdownDescription:
                (with_superbol_toml_note "Default source reference-format.")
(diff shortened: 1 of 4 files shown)


src/lsp/superbol_free_lib/vscode_extension.ml
```diff
@@ -414,13 +415,13 @@ let contributes =
                            numerics)."
              ~order:31;
 
-           Manifest.PROPERTY.null_string "superbol.debugger.libcob-path"
+           Manifest.PROPERTY.null_string "superbol.debugger.libcobPath"
```
nberth, line 418: @engboris If you changed this then yes I think that `settings.ts` needs to be updated. But note it's in `superbol-vscode-debug`, which is another repository (and submodule for `superbol-vscode-oss`).

engboris:
Tests have been done for:
- `source-format` (by changing the workspace config and see effect on source code);
- `cobc` (by looking at the path used by the `F5` command);
- `superbol.toml` which outputs what is expected.

I probably missed something and I'm not sure how to properly test `libcob-path` and `gdb-path`.

src/lsp/superbol_free_lib/vscode_extension.ml
engboris: What is the purpose of this file?

nberth: What do you mean? What file?

engboris: Ok I think I figured out myself. I was thinking that the internal representation of configs used dashes so I didn't see why getters for debug settings had to be changed but it's about the client-side (VSCode extension) which uses CamelCase.

GitMensch:
You can at least test them when building it locally (needs to be pushed to the sub-repo later) and running it.

`gdb-path` is the debugger binary that is started (just set to an invalid entry, start, get an error, set to full path, start, have debugging, unset the path [so it is searched for in PATH], start debugging)
`display-variable-attributes` -> check (was it with hover?) how the variables are shown in the debugger view when this is (not) active.

nberth:
Also: https://github.com/OCamlPro/superbol-vscode-debug/pull/12 needs to be merged first, and the submodule update be included in the current PR.

PR #425, opened 2025-05-28
Fix lookup of task configurations

### LINKED ISSUE #421: bug: task definition does not use cobcPath
even if set in the task config it uses the superbol.cobcPath configuration, and this defaults to `cobc`

nberth:
Fixes #421.


engboris:
Looks like I missed some relevant occurrences of config names in my last PR :/

PR #428, opened 2025-05-28
switch problemMatcher patterns from line to locations

GitMensch:
which should highlight the complete line and reference not only column 1 in the problem pane

warning: this is untested, it just worked when I tested it locally with my own problemMatcher definitions (and got that hint from a vscode dev); obviously "some day" we should emit columns from cobc as well - but if we do it the gcc way "location" should handle those fine (we likely will need to adjust the regex, when the time comes)

... maybe I try the CI build of this PR some day to test the result, but there's possibly someone else faster :-)

### DIFF
--- a/package.json
+++ b/package.json
@@ -507,33 +507,33 @@
         "name": "gnucobol",
         "regexp": "^(.*): ?(\\d+): (error|warning): ([^[]*)(\\[(.*)\\])?$",
         "file": 1,
-        "location": 2,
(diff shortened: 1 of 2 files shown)


nberth:
> which should highlight the complete line and reference not only column 1 in the problem pane

Oh I see. Great idea indeed (I guess the difference between "line" and "location" is hidden in plain sight in vscode's api documentation.

PR #452, opened 2025-11-05
WIP: Update debug extension

emilienlemaire:


src/lsp/superbol_free_lib/vscode_extension.ml
```diff
@@ -250,6 +250,20 @@ let contributes =
...
+                  "type": "superbol-gdb",
+                  "request": "launch",
+                  "preLaunchTask": "SuperBOL: build (debug)",
+                  "target": "${file}",
+                  "arguments": "",
+                  "cwd": "${workspaceFolder}",
+                  "group": [],
+                  "coverage": true,
+                  "verbose": false,
+                  "gdbtty": true
```
nberth: To me, this is very annoying to have a terminal widow up on every debug session. If that's what people typically get on Windows then I'd be to find a way to enable that by default on that platform only.

nberth:
Actually I fail to see what the proposed changes are doing; isn't it just mimicking what's already coded in `resolveDebugConfiguration` of the debug extension (`src/extension.ts`)?

PR #453, opened 2025-11-07
Fix handling of program name with dashes, and files outside of a workspace

### LINKED ISSUE #355: superbol-debug: allow ad-hoc debugging "execute and debug" for files outside of a workspace
Using that command currently comes up with "install extension for COBOL"; the current default one which includes `"cwd": "${workspaceFolder}"` cannot work as this variable is not set for non-workspace windows.

### LINKED ISSUE #451: Debugger fails to display variables in certain GnuCOBOL source files
I have a problem displaying variables in the SuperBOL Studio Debugger.

The problem depends on the COBOL source files.

The debugger shows variables and their values in the example programs logreader.cob and logsort.cob.

However, the debugger does not show variables in the example programs simple-hello.cob and bubblesort.cob.

It seems that simpler programs have more problems than more complex ones, such as eliza.cob (Arnold Trembley’s version) and customer.cob (from the cobgdb demo).

I have documented this issue on Ubuntu 25.10 (see the environment directory), but it also exists on Windows 11 with SuperBOL-GnuCOBOL (June 2024) and under WSL.

Another test with cobgdb runs well and displays variable values correctly.

Screenshots and source code are attached in the ZIP file.

Source is also available on https://github.com/mds711/gnucobol-debug-tests

Could you please investigate why simple-hello.cob and bubblesort.cob fail to display variable values during debugging in SuperBOL Studio?

Any guidance or fixes would be greatly appreciated.

P.S.: I really like SuperBOL Studio — it’s a great extension for GnuCOBOL with VS Code. Thank you very much for your work!

[gc-debug-ticket.zip](https://github.com/user-attachments/files/23340750/gc-debug-ticket.zip)

Image

nberth:
Fixes #451
Fixes #355

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## [_] Next release
 
+### Added
+- Ability to launch a debug session when only a single file is opened (without a workspace) [#453](https://github.com/OCamlPro/superbol-studio-oss/pull/453) (fix for [issue #355](https://github.com/OCamlPro/superbol-studio-oss/issues/355))
+
+### Fixed
+- Handling of program names with dashes by the debugger [#453](https://github.com/OCamlPro/superbol-studio-oss/pull/453) (fix for [issue #451](https://github.com/OCamlPro/superbol-studio-oss/issues/451))
+
 
 ## [0.2.1] Bugfix release (2025-09-23)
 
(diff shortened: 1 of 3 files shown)


emilienlemaire (approved):
If the limitation on the `gdbtty` are OK, then Ok to merge, otherwise please revert the default `gdbtty` value

PR #468, opened 2025-12-16
Fix fallback when `cobcPath` is not provided in task definitions

### LINKED ISSUE #467: Extension setting `superbol.cobcPath` is not used as fallback when `cobcPath` is not provided in task definitions
When trying the extension during OSXP I've recognized that the path for `cobc` was not used from the extension setting and therefore needed to be specified _additionally_ in the task definition; Can you please check if this is still the case and if yes tackle that?
If it is not set in the task then it should use the extension setting, if that is empty as well then it should use `cobc`; the _internal_ definitions (used if not configured manually) should not set this, to again fall back to the extension setting which should fall back to cobc.

_Originally posted by @GitMensch in https://github.com/OCamlPro/superbol-studio-oss/issues/466#issuecomment-3660540256_

nberth:
Fixes #467

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Basic language definition for toggling line comments [#463](https://github.com/OCamlPro/superbol-studio-oss/pull/463)
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


nberth:
> Is there a way for automatically testing these kinds of things?

There is probably a way. I managed to get some work done recently to re-enable, fix, and expand the tests of the debugger extension, and make them run in the CI for that separate repository. A first step would be to run the same tests as the debugger extension via the whole SuperBOL Studio extension itself (although those automated tests do not involve workspace manipulations yet). Those tests would also require installing GnuCOBOL and other dev tools.

PR #492, opened 2026-02-05
Add binding for the showOpenDialog function of Vscode.Window

engboris:
A binding for the `showOpenDialog` function of `Vscode.Window` has been added. It uses a `OpenDialogOptions` module for its arguments. I took inspiration from the `MessageOptions` module (just above).


nberth (approved):
Perfect, thanks!

PR #499, opened 2026-02-09
Support hexadecimal litterals

engboris:
Support for hexadecimal literals as in

```cob
       78 AAA             value h"00000001".
```

Support both `H` and `h` following already existing `Hex` variant (originally for `X`). Added a `HexH` variant for better reprinting after suggestion for the reviewer. The code added follow the code of the existing `Hex` variant.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### Added
 - Support for Hexadecimal literals using 'H' and 'h' prefix [#499](https://github.com/OCamlPro/superbol-studio-oss/pull/499)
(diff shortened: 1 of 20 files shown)


nberth:
You may also need to add support for those in `cobol_preproc/text_types.ml` (`literal_kind`, where the `h` onces are I think synonym of `x` — with an `HexH` maybe for proper re-printing) and `cobol_preproc/src_lexer.mll` (`alphanum_lit_new`).

nberth (approved):
Thanks!

PR #512, opened 2026-02-10
Better handling of spaces before MF compiler directives

emilienlemaire:


nberth (approved):
Looks ok (malformed compiler directives test is very artificial so we can keep it untouched).

PR #515, opened 2026-02-10
Support more compiler directive syntaxes

GBertholon:
Namely `$SET INTLEVEL`, `$SET NSYMBOL`, `$SET SIGN` and `$IF x SET`.

All of them generate warnings further down the chain since we do not interpret them yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [_] Next release
 
 ### Added
+- Support more compiler directive syntaxes [#515](https://github.com/OCamlPro/superbol-studio-oss/pull/515)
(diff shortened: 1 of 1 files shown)


src/lsp/cobol_preproc/compdir_grammar.mly
```diff
@@ -144,32 +145,44 @@ let set_directive :=
 let set :=
   | ~ = loc(set_operand); <Set>
 
+let set_value :=
+  | ~ = loc(ALPHANUM);             <Alphanum>
+  | LPAR; ~ = loc(FIXEDLIT); RPAR; <Numeric>
+
+let set_string_value :=
```
nberth:
```suggestion
let string_value :=
```
Maybe…

nberth (approved):
Great. Ok to merge after new entry in changelog, and optionally addressing the two comments.

PR #516, opened 2026-02-10
Add a platform record to virtualize some calls and heuristics

lefessan:


src/lsp/cobol_lsp/lsp_platform.ml
```diff
@@ -0,0 +1,14 @@
...
+(*  Copyright (c) 2022-2023 OCamlPro SAS                                  *)
+(*                                                                        *)
+(* All rights reserved.                                                   *)
+(* This source code is licensed under the GNU Affero General Public       *)
+(* License version 3 found in the LICENSE.md file in the root directory   *)
+(* of this source tree.                                                   *)
+(*                                                                        *)
+(**************************************************************************)
+
+let record = Cobol_common.Platform.default
```
lefessan: We should probably migrate the definition of `Platform.default` here. The idea is that `default` should not appear in `Cobol_common`, because it creates dependencies that we actually want to remove. Such dependencies are ok to create at the application level, for example here.

PR #524, opened 2026-02-11
Internalize JS dependencies for CFG feature

engboris:
The dependencies have been downloaded and included as "vendor assets". This will enable the CFG feature to work offline or with proxy.


engboris:
> Thanks! This likely addresses the need for Internet access issue. Before merging we need to check whether we need to include a LICENCES file in `assets/vendor`, or a related note somewhere else.

Done.

nberth (approved):

PR #526, opened 2026-02-12
Support INDEXED BY and sort direction in any order in OCCURS clause

engboris:
In the `OCCURS` clause, the `INDEXED BY` clause and sort directions (`ASCENDING KEY IS` and `DESCENDING KEY IS`) can be used in any order but only one order was supported.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## [_] Next release
 
 ### Added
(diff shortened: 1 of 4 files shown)


nberth (approved):
LGTM. ~~I guess you can enlist this PR in the generic-enough "Various minor improvements to the COBOL grammar" CHANGELOG entry (already existing).~~

PR #527, opened 2026-02-12
Optional `program-id` paragraph

emilienlemaire:

PR #530, opened 2026-02-13
Handle tabs in input

emilienlemaire:
Based on #527

### DIFF
--- a/src/lsp/cobol_common/srcloc.ml
+++ b/src/lsp/cobol_common/srcloc.ml
@@ -363,8 +363,9 @@ let pp_file_loc ppf ((file, pos1, pos2): raw_loc) =
 
 (** Retrieve the character position of a column number in a line that was
       tab expanded *)
-let original_col_of_expanded ?(tab_stop = 8) ~platform ~fname ~line expanded_col =
+let original_col_of_expanded ~platform ~fname ~line expanded_col =
   try
+    let tab_stop = platform.tab_stop in
     let lines = retrieve_file_lines ~platform fname in
     let original_line = lines.(line - 1) in
     let n = String.length original_line in
@@ -433,9 +434,10 @@ let tab_underline_to_end ?(tab_stop = 8) ~lo line =
   Buffer.contents buf
 
 (** Note this should always end with a newline character *)
-let pp_raw_loc: ?tab_stop:int -> ?platform:platform -> raw_loc Pretty.printer =
+let pp_raw_loc: ?platform:platform -> raw_loc Pretty.printer =
   let b = lazy (Buffer.create 1000) in
-  let find_source ~tab_stop ~platform (file, pos1, pos2) =
+  let find_source ~platform (file, pos1, pos2) =
+    let tab_stop = platform.tab_stop in
     let line1 = fst pos1 in
     let line2 = fst pos2 in
     let col1 = snd pos1 in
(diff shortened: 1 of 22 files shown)


src/lsp/cobol_preproc/src_reader.ml
```diff
@@ -242,10 +242,54 @@ let make make_lexing ?filename ~source_format input =
   Option.iter (Lexing.set_filename lexbuf) filename;
   Plx (Src_lexing.init_state source_format, lexbuf)
 
+let from_channel_expanding_tabs ?with_positions ?(tab_stop = 8) ic : Lexing.lexbuf =
```
nberth: To be placed at the end of `Src_lexing` for now. Or a new `Src_rewrtiting` module (with `expand_tabs`) maybe.

test/output-tests/used_binaries.expected
```diff
@@ -48,12 +48,12 @@ Considering: import/gnucobol/tests/testsuite.src/used_binaries.at:946:0
...
 Considering: import/gnucobol/tests/testsuite.src/used_binaries.at:991:0
 Considering: import/gnucobol/tests/testsuite.src/used_binaries.at:1024:0
-used_binaries.at-1024-progprep.cob:5.36-5.37:
+used_binaries.at-1024-progprep.cob:5.47-5.48:
    2          IDENTIFICATION   DIVISION.
    3          PROGRAM-ID.      prog.
    4          DATA             DIVISION.
    5 >        WORKING-STORAGE  SECTION.	  	#
-----                                       ^
+----
```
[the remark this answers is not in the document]
nberth: The actual source file contains tab characters, and that's the one we feed to the parser (not the prepared one). I think we should still get a proper positioning of the marker (even is a bit off in the diff maybe), as that's the point of this PR.

src/lsp/cobol_preproc/src_lexing.ml
```diff
@@ -675,3 +675,4 @@ let separator' ~char ~k = free_text (free_separator ~char) ~k
 let alphanum_lit' ~k = free_text free_alphanum_lit ~k
 
 (* --- *)
+
```
nberth, line 742: Please avoid unnecessary space changes (in `src_lexing.mli` as well).

src/lsp/cobol_common/srcloc.ml
```diff
@@ -349,16 +349,87 @@ type raw_loc = string * (int * int) * (int * int)
...
+        Buffer.add_char buf (if !vcol + dv > lo then '^' else ' ')
+      done;
+      vcol := !vcol + w )
+    line;
+  Buffer.add_char buf (if !vcol > lo then '^' else ' ');
+  Buffer.contents buf
+
 (** Note this should always end with a newline character *)
-let pp_raw_loc: raw_loc Pretty.printer =
+let pp_raw_loc ?(tab_stop = 8) () : raw_loc Pretty.printer =
```
nberth: This additional argument and the impact of the PR makes me wonder whether early source-rewriting is the proper solution (as tab-stops should be attached to locations, and should become a configurable list in the future…)… I think we should rethink a bit more on that.

nberth (changes requested):
I'm not so found of the current approach, as I think the rewriting is quite intrusive. I think a cleaner solution would be to implicitly account for shifted columns into `src_lexing` (with a new `tab` function that works similarly to other functions of this module — to simplify we can assume there are no tabulation characters in picture strings: the main issue is about proper computation of columns in fixed format, the other aspects will be easy to address afterwards). As a aside, the tab stops configuration should not belong to the platform, but to the configuration of the source lexer.

src/lsp/cobol_preproc/src_lexing.ml
emilienlemaire: I think that's a byproduct of my rebasing

nberth:
Subsumed by #591

PR #532, opened 2026-02-13
Handle OSVS parentheses in abbreviated combined relation conditions

GBertholon:
This commit adds support for OSVS condition syntax such as: 
```cob
a = (1 OR 2)
a > b OR (c AND d)
a > (b OR c) AND d
a (= b OR > c)
a = b AND (> c OR < d)
```

It also changes the internal representation of abreviated combined relation conditions to include a correct representation of implicit associativity.

To avoid grammar ambiguity with subscripts we need to adjust the lexer to look ahead of '(' characters and tell if the next element is a relational operator.

This commit removes one example of condition parsing where MF and GnuCOBOL do not agree:
```cob
A = "a" OR (NOT B)
```

The new abbreviation expansion function now follows GnuCOBOL's behaviour on this example while the previous one was trying to follow MF.

### DIFF
--- a/src/lsp/cobol_ptree/terms_helpers.ml
+++ b/src/lsp/cobol_ptree/terms_helpers.ml
@@ -26,6 +26,17 @@ type abbrev_condition_expansion_state =
   | AfterRelOp of bool * expression * relop
   | AfterNonAbbrev
 
+(** [AbbrevMissingSubjectAfterNonAbbrev] is raised by [expand_abbrev_cond] on
+    invalid conditions such as [a < b OR c IS POSITIVE OR > d] where there is no
+    rule to expand [> d] accoding to the COBOL standard. *)
+exception AbbrevMissingSubjectAfterNonAbbrev
+
+(** [AbbrevMissingRelOpAfterSubject] is raised by [expand_abbrev_cond] on invalid
+    conditions where there is no [AbbrevRelOp] in leftmost position after an
+    [AbbrevSubject]. Normally such cases are forbidden by parsing rules so this
+    exception is an internal error. *)
+exception AbbrevMissingRelOpAfterSubject
+
 (** [expand_every_abbrev_cond cond] recursively substitutes every abbreviated
     combined relation condition from [cond] by an equivalent non-abbreviated
     condition (with abbreviated relations replaced with binary relations). *)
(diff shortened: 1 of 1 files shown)


src/lsp/cobol_parser/text_tokenizer.ml
```diff
@@ -73,6 +73,12 @@ let preproc_n_combine_tokens ~intrinsics_enabled ~source_format =
         | i -> cons x acc (tl p, hd l +@+ hd (tl l) :: tl (tl l)) (i - 1)
       in
       cons x acc (p, l) y
+    and replace_n x y =
```
nberth: Shouldn't this be named `replace_after_n` or similar?

test/cobol_parsing/test_combined_relations_parsing.ml
```diff
@@ -102,15 +104,13 @@ let test_conditions =
        OR; DIGITS "2"]                       ((ae ==. one) ||. (ae ==. two));
     chk "A = 1 OR 2 OR 2 = B"
       [a; EQ; DIGITS "1"; OR;
-       DIGITS "2"; OR; DIGITS "2"; EQ; b]    ((ae ==. one) ||.
-                                              ((ae ==. two) ||. (two ==. be)));
+       DIGITS "2"; OR; DIGITS "2"; EQ; b]    (((ae ==. one) ||. (ae ==. two)) ||. (two ==. be));
```
nberth:
```suggestion
       DIGITS "2"; OR; DIGITS "2"; EQ; b]    (((ae ==. one) ||. (ae ==. two)) ||.
                                              (two ==. be));
```

src/lsp/cobol_ptree/terms_helpers.ml
```diff
@@ -41,43 +49,44 @@ let rec expand_every_abbrev_cond
...
+        neg_condition ~neg c, sr
+    | AbbrevNot a, sr ->
+        let c, sr = disambiguate a sr in
+        Not c, sr
+    | AbbrevOther c, _ ->
+        expand_every_abbrev_cond c, AfterNonAbbrev
+    | AbbrevComb (a1, logop, a2), sr ->
+        let c1, sr = disambiguate a1 sr in
+        let c2, sr = disambiguate a2 sr in
+        Logop (c1, logop, c2), sr
```
nberth: Could we either:
- convince ourselves the failure cases cannot happen on any input to the grammar;
- find a way to deal with these cases more gracefuly?

nberth (commented):
Wonderful, many thanks for tackling this hard problem. Only issue I see is about the  few `failwith` that may occur during parsing, and that only a much more careful review may judge safe-enough (later).

PR #534, opened 2026-02-15
Attempt at fixing Text_tokenizer.tokenize_after

lefessan:
This function is called repeatedly when entering/leaving grammar contexts with Enabled_keywords/Disabled_keywords, but everytime, it updates the whole list of tokens, most of the time redoing/undoing what it had done before.

In this attempt, in the case of Enabled_keywords/Disabled_keywords, we don't update the list of tokens, instead we keep the list of enabled keywords and translate them on the fly when the next token is asked.

This version has a 10x speedup on problematic file.

HOWEVER, it does pass the testsuite, and, unfortunately, the testsuite lacks documentation on how to fix problems using it (i.e. how do you run a specific test in isolation ?).

I left a lot of debug stuff in the code, especially, it currently compares its output with the former implementation and fails if they disagree. Should ease testing...

### DIFF
--- a/Makefile.header
+++ b/Makefile.header
@@ -82,6 +82,10 @@ test-syntax-promote:
 	-$(MAKE) test-syntax	# To avoid promoting past tests
 	\! opam exec -- autofonce 2>/dev/null || opam exec -- autofonce promote -t syntax --apply
 
+.PHONY: test-deps
+test-deps:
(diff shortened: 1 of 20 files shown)


lefessan:
There may be other ways… yet, this particular change yielded a 10x speedup on some files (3s -> 0.3s), so it seems it must be done for these files even if other speedups are found for other files.

PR #575, opened 2026-04-09
Fix size shown for data items with yet-unsupported usage or erroneous definitions

nberth:
Additionally:
- show sizes in bytes when relevant;
- indicate when an item is a "variable" ODO.

### DIFF
--- a/src/lsp/cobol_lsp/lsp_data_info_printer.ml
+++ b/src/lsp/cobol_lsp/lsp_data_info_printer.ml
@@ -16,7 +16,19 @@ open Cobol_data.Types
 open Cobol_common.Srcloc.TYPES
 open Cobol_common.Srcloc.INFIX
 
-let pp_size = Fmt.(any "Size: " ++ Cobol_data.Memory.pp_size ++ any " bits")
+let pp_readable_size ppf size =
+  try
+    let bits = Cobol_data.Memory.as_bits size in
+    if Int.rem bits 8 = 0 then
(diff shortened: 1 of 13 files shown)


src/lsp/cobol_lsp/lsp_data_info_printer.ml
```diff
@@ -221,18 +248,19 @@ let pp_renamed_item_layout: renamed_item_layout Pretty.printer = fun ppf x ->
...
+    end
     ++ any "\n\n"
     ++ const pp_renamed_item_layout r.renaming_layout
     ++ (match r.renaming_layout with
-    | Renamed_struct _ -> any "  \n" ++ const pp_size r.renaming_size
-    | _ -> nop) )
-  ppf r
+        | Renamed_struct _ -> any "  \n" ++ const pp_size r.renaming_size
+        | _ -> nop)
+  end ppf r
```
emilienlemaire, line 265: This is hardly readable...
Was this reformatting necessary?

emilienlemaire (approved):
Looks good to me, but one function reformatting

src/lsp/cobol_lsp/lsp_data_info_printer.ml
nberth: Rewriting it this way was the fastest way for me to introduce the guards on printing combinators in such a way that the whole would still indent properly. In any case both formats are a pain to edit; yet for me the new one is more readable.

PR #576, opened 2026-04-16
Documentation and renaming to reduce confusion in the typed COBOL AST API

GBertholon:
This PR tries to add missing documentation on the typed COBOL AST.

Currently this is a draft with open questions where I need some input to produce a useful and missing documentation.

I also suggest some very minor refactorings that I am ready to perform.

### DIFF
--- a/src/lsp/cobol_data/data_types.ml
+++ b/src/lsp/cobol_data/data_types.ml
@@ -77,20 +77,23 @@ and item_definitions = item_definition with_loc nel
 and item_redefinitions = item_definition with_loc list
 
 and item_definition =
-  | Field of field_definition
-  | Table of table_definition
+  | Field of field_definition (** for data items without OCCURS clause *)
+  | Table of table_definition (** for data items with an OCCURS clause *)
 
 and field_definition =
   {
     field_qualname: Cobol_ptree.qualname with_loc option;
-    field_redefines: Cobol_ptree.qualname with_loc option;      (* redef only *)
     field_leading_ranges: table_range list;
     field_offset: Data_memory.offset;         (** offset w.r.t record address *)
     field_size: Data_memory.size;
     field_layout: field_layout;
     field_length_variability: length_variability;
-    field_conditions: condition_names;
-    field_redefinitions: item_redefinitions;
+    field_conditions: condition_names; (** Named conditions on the value of this field. *)
+    field_redefines: Cobol_ptree.qualname with_loc option; (** 
+      Set iff this field is a redefinition. 
+      In that case this field appears inside item_redefinitions of the item it redefines. 
+      Later, we may create instead a item_redefinition type. *)
+    field_redefinitions: item_redefinitions; (** List of alternative definitions for this field *)
     field_has_definition_issues: bool;
   }
 
@@ -112,8 +115,10 @@ and table_definition =
     table_size: Data_memory.size;
     table_range: table_range;
     table_init_values: Cobol_ptree.literal with_loc list;     (* list for now *)
-    table_redefines: Cobol_ptree.qualname with_loc option;    (* redef only *)
-    table_redefinitions: item_redefinitions;
+    table_redefines: Cobol_ptree.qualname with_loc option; (* same as [field_redefines] but for tables *)
+    table_redefinitions: item_redefinitions; (** 
+      List of alternative definitions for the full table. 
+      Note that by default the typechecker generates a warning on table redefinition. *)
     table_has_definition_issues: bool;
   }
 and table_range =
@@ -144,6 +149,7 @@ and dynamic_span =
     occurs_dynamic_initialized: bool with_loc;
   }
 
+(** Named conditions a.k.a. level 88 items *)
 and condition_names = condition_name with_loc list
 and condition_name =
   {
@@ -190,6 +196,8 @@ and renamed_item_layout =
 (*     const_layout: const_layout; *)
 (*   } *)
 
+(** [data_definition] provides a direct access to the pair of each item and associated record items.
+    It is used for instance to retrieve data item information from its name. *)
 type data_definition =
   | Data_field of
       {
--- a/src/lsp/cobol_typeck/typeck_outputs.ml
+++ b/src/lsp/cobol_typeck/typeck_outputs.ml
@@ -16,6 +16,7 @@ open Cobol_common.Srcloc.INFIX
 open Cobol_common.Exec_block.TYPES
 module LIST = Cobol_common.Basics.LIST
 
+(** Contains the locations of syntactic occurences for each qualified name *)
 type qualrefmap = srcloc list Cobol_unit.Qual.MAP.t
 type references_in_unit =
   {
@@ -38,9 +39,12 @@ type artifacts =
 
 type outputs =
   {
-    ptree: Cobol_ptree.compilation_group;
-    group: Cobol_unit.Types.group;
-    artifacts: artifacts;
+    ptree: Cobol_ptree.compilation_group; (** 
+      Parse tree (AST without typing information) copied in the typechecker output 
+      to ease its direct use by further passes in the analysis pipeline and queries 
+      in the LSP server. *)
+    group: Cobol_unit.Types.group; (** The typed AST view as an unordered set of cobol_unit *)
+    artifacts: artifacts; (** Other typing artifacts *)
   }
 type t = outputs
 
--- a/src/lsp/cobol_unit/unit_resolver_map.ml
+++ b/src/lsp/cobol_unit/unit_resolver_map.ml
@@ -22,7 +22,7 @@ module NEL = Cobol_common.Basics.NEL
 
 module TYPES = struct
 
-  type 'a qualmap =
+  type 'a resolver_map =
     {
       map: 'a node StrMap.t;
       indirect_quals: StrSet.t;            (* indirect qualifiers used in map *)
@@ -31,7 +31,7 @@ module TYPES = struct
   and 'a node =
     | Exact of {
         binding: 'a binding option;
-        refined: 'a qualmap;
+        refined: 'a resolver_map;
       }
     | Cut of {
         binding: 'a binding;
@@ -49,7 +49,14 @@ module TYPES = struct
 end
 include TYPES
 
-type +'a t = 'a qualmap
+type +'a t = 'a resolver_map
+
+let () =
+  Printexc.register_printer (function
+      | Ambiguous (lazy qns) ->
+          Some (Pretty.to_string "Ambiguous reference found: %a"
+                  (NEL.pp Unit_qual.pp) qns)
+      | _ -> None)
 
 (* --- *)
 
@@ -84,12 +91,12 @@ let pp_binding pp_value ppf { value; full_qn; _ } =
     pp_qualname full_qn
     pp_value value
 
-let pp_qualmap pp_value ppf map =
+let pp pp_value ppf map =
   Pretty.list ~fopen:"{@;<1 2>" ~fsep:"@;<1 2>" ~fclose:"@;}" ~fempty:"{}"
     (pp_binding pp_value) ppf
     (bindings map)
 
-let pp_qualmap_struct pp_value ppf map =
+let pp_struct pp_value ppf map =
   let rec pp_map ppf { map; _ } =
     Pretty.list ~fopen:"{|@;<1 2>" ~fsep:"@;<1 2>" ~fclose:"@;|}" ~fempty:"{/}"
       (fun ppf (key, node) -> Pretty.print ppf "@[<hv 2>%S:@ %a@]" key pp_node node)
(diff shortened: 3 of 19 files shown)


src/lsp/cobol_data/data_types.ml
```diff
@@ -83,14 +83,14 @@ and item_definition =
...
     field_leading_ranges: table_range list;
     field_offset: Data_memory.offset;         (** offset w.r.t record address *)
     field_size: Data_memory.size;
     field_layout: field_layout;
     field_length_variability: length_variability;
-    field_conditions: condition_names;
-    field_redefinitions: item_redefinitions;
+    field_conditions: condition_names; (** Named conditions on the value of this field. *)
+    field_redefines: Cobol_ptree.qualname with_loc option;      (** set iff this field is a redefinition. In that case this field must appear only inside item_redefinitions of the item it redefines. TODO: remove and extend item_redefintions type instead? *)
+    field_redefinitions: item_redefinitions;  (** list alternative definitions for this field *)
```
nberth: I like the idea of extending `item_redefinitions` instead, probably with a record like `{ item_redefinitions_redefine: Cobol_ptree.qualname with_loc; item_redefinitions: item_definition wiith_loc nel; }`, and then having `field_redefinitions: item_redefinitions option;` or something like that.

src/lsp/cobol_typeck/typeck_outputs.ml
```diff
@@ -38,9 +38,9 @@ type artifacts =
 
 type outputs =
   {
-    ptree: Cobol_ptree.compilation_group;
-    group: Cobol_unit.Types.group;
-    artifacts: artifacts;
+    ptree: Cobol_ptree.compilation_group; (* Parse tree to access information missing in the typed AST. Is this field supposed to be stable? Should it only be used to access things that are inaccessible from cobol_unit ? *)
+    group: Cobol_unit.Types.group; (* The typed AST view as an unordered set of cobol_unit *)
+    artifacts: artifacts; (* Other typing artifacts *)
```
nberth: `ptree` is left here to ease its direct use by further passes in the analysis pipeline and queries in the LSP server (in which some position-based queries need to be performed on the parse-tree instead of the typed AST). Its presence here is not mandatory so it could be removed.

src/lsp/cobol_unit/unit_types.ml
```diff
@@ -52,9 +52,12 @@ type unit_config =
...
 (* data items *)
 
+(** data_definitions exposes two views on the records in DATA DIVISION: 
+   - [data_items] contains all the fields with direct access by name.
+   - [data_records] is a list of record trees where each element represent a record root (typically at level 01) *)
 type data_definitions =
   {
-    data_items: Cobol_data.Types.data_definition named_n_ordered;
+    data_items: Cobol_data.Types.data_definition named_n_ordered; (* FIXME: Is the ordered part useful here ? Isn't it already exposed in a more structured way by data_records... *)
     data_records: Cobol_data.Types.record list;
```
nberth, line 64: Here it makes sense to keep the ordered alternative as some data items are anonymous (fillers) and therefore do not belong to the resolution map. Additionally the `data_definition` brings more info than `item_definitions` themselves, that are the types accessible from the `record` descriptions.

src/lsp/cobol_unit/unit_types.ml
```diff
@@ -52,9 +52,12 @@ type unit_config =
 
 (* data items *)
 
+(** data_definitions exposes two views on the records in DATA DIVISION: 
+   - [data_items] contains all the fields with direct access by name.
```
nberth: Note it's not "fields" but "items": an item is either a field or a table, and a table is always anonymous.
So `data_items` "contains all data items with direct access by name or ordered access."

src/lsp/cobol_unit/unit_types.ml
```diff
@@ -52,9 +52,12 @@ type unit_config =
...
+   - [data_records] is a list of record trees where each element represent a record root (typically at level 01) *)
```
nberth:
```suggestion
   - [data_records] is a list of record trees where each element represents a record definition (typically a level 01 or 77 data item) *)
```

nberth (commented):
Interesting investigations. Some of my comments describe changes that could also be included in this PR. In addition, please fit the comments within 80 characters (on the lines that follow field definitions I think).

src/lsp/cobol_unit/unit_types.ml
GBertholon: Which information is only accessible through `data_definition` ? I cannot find any myself...

src/lsp/cobol_typeck/typeck_outputs.ml
GBertholon: Since it is for easing direct access with a pipeline setup it is obviously not mandatory but the question is more: do we want Cobol_typeck clients to see and use this or is it an internal quirk for LSP ?

nberth: Lets say for now this is only for the LSP server, so it can probably be removed without too much hassle.

src/lsp/cobol_unit/unit_types.ml
nberth: `data_definition` provides a direct access to the pair of each item and associated records, in addition to typing info (e.g. for indexes, data renaming) that is more readable and conveniently accessed than via an in-depth exploration of the corresponding record definitions. Given that work is soon going to start on TramaBOL, I am not in favor of removing it unless it is replaced with a view that provides similar information; we can reconsider this if it appears we don't need this kind of access in TramaBOL.

src/lsp/cobol_data/data_types.ml
GBertholon: I am thinking more of:
```ocaml
type item_redefinition =
  {
    redefinition_redefines: Cobol_ptree.qualname with_loc; (** Name mentionned after the REDEFINES keyword *)
    redefinition_item: item_definition; (** How is it redefined *)
  }
and item_redefinitions = item_redefinition with_loc list
```
since we may need the location of each qualname after the REDEFINES keyword.

I have experimented around this and it seems like it imposes too much refactoring for this simple PR.
For example we also would need a new `data_definition` kind to be able to find the redefinition from its name.

src/lsp/cobol_unit/unit_types.ml
GBertholon: So you mean a view that turns a tree into ordered `data_definition` ? I guess this shouldn't be hard but I have no examples in mind where I would prefer an ordered list of `data_definition`s instead of a tree that I can visit. Since the change is minimal let's indeed reconsider later.

src/lsp/cobol_unit/unit_types.ml
GBertholon: OK. The word item is indeed used by the COBOL standard but is hard to understand from a non COBOList.
I try to use item or data item everywhere but in some places such as here I will recall that this corresponds to fields + tables.
In this vein, I guess I should explain that tables are data items with OCCURS.

src/lsp/cobol_data/data_types.ml
nberth: I am not sure a "data definition" needs to directly inform that an item is a redefinition (which is what you seem to want to provide): I think every redefinition case is already covered by existing constructors. Similarly I don't see a use for a dedicated lookup of a particular named redefinition either.

src/lsp/cobol_unit/unit_types.ml
nberth: Yes as one would say in COBOL a data item with an OCCURS clause implicitly induces a table.

src/lsp/cobol_data/data_types.ml
GBertholon: My attempt at refactoring seems to show that we are currently using name lookup that gives data_definition to generate hover information in the LSP and this information includes whether the field is a redefinition or not, hence my comment.

src/lsp/cobol_typeck/typeck_outputs.ml
```diff
@@ -28,6 +28,9 @@ type references_in_unit =
          - a indirect reference to A
       In the future we may need/want to split those 2 kind of references
      *)
+    (* TODO: Can we really have a one fits all notion of references_in_unit?
+       Is this only supposed to be references displayed in the LSP?
+       In that case shouldn't we drop indirect references altogether? *)
```
nberth: I think the indirect ones matter at least to support symbol renaming requests (e.g. renaming `A` in `B IN A`) . If keeping those is an issue and we find another way to deal with such requests we could drop indirect refs, but I wouldn't consider this an issue for now.

nberth (approved):
Thanks! Could be merged unless there is a point I didn't see for which you want to discuss more.

src/lsp/cobol_typeck/typeck_outputs.ml
GBertholon: OK so to clarify things, I guess we should write that these qualrefmap contain all syntactic occurrence of each name.

GBertholon:
Before considering this task done, I would still like to include some documentation / renaming to deal with the situation of Qualmap:
- First, we have a type `Cobol_unit.Qual.MAP` which is the `Map` functor applied on type `qualname`
- Then, we have `Unit_qualmap` reexported as `Cobol_unit.Qualmap` that can do name resolution
- Finally we have `Cobol_data.Qualmap` that looks like `Cobol_unit.Qualmap` from the outside but with a different implementation.

To add more confusion, most `qualname`s are in fact syntactic optionally qualified names that should be used in `Cobol_unit.Qualmap` to be resolved, but some are fully qualified and can be used meaningfully with `Cobol_unit.Qual.MAP`.

GBertholon:
I found that the file `src/lsp/cobol_data/qualmap.ml{,i}` is not exported as `Cobol_data.Qualmap` but as `Cobol_data.OLD.Qualmap` so it is probably a deprecated API. The only problem here for me is that the filename does not say that this is the old API. Is there a good reason to keep those old APIs though?

nberth:
> I found that the file `src/lsp/cobol_data/qualmap.ml{,i}` is not exported as `Cobol_data.Qualmap` but as `Cobol_data.OLD.Qualmap` so it is probably a deprecated API. The only problem here for me is that the filename does not say that this is the old API. Is there a good reason to keep those old APIs though?

Yes I remembered that aspect over lunch and wanted to comment on that. There is somewhere some old typeck0-checing code that an external tool is using (maybe also in `Cobol_typeck.OLD` or `old_cobol_typeck` or similar). I'd be greatly in favor of removing this code.

nberth:
> Before considering this task done, I would still like to include some documentation / renaming to deal with the situation of Qualmap:
> 
> * First, we have a type `Cobol_unit.Qual.MAP` which is the `Map` functor applied on type `qualname`
> * Then, we have `Unit_qualmap` reexported as `Cobol_unit.Qualmap` that can do name resolution
> * Finally we have `Cobol_data.Qualmap` that looks like `Cobol_unit.Qualmap` from the outside but with a different implementation.
> 
> To add more confusion, most `qualname`s are in fact syntactic optionally qualified names that should be used in `Cobol_unit.Qualmap` to be resolved, but some are fully qualified and can be used meaningfully with `Cobol_unit.Qual.MAP`.

To remove some confusion I'd favor renaming `Cobol_unit.Qualmap` into a name that you suggested like `Cobol_unit.Resolution_map`. Moreover, is  `Cobol_unit.Qual.MAP` used at all?

src/lsp/cobol_typeck/typeck_outputs.ml
nberth: (I seem to have noted a comment where you somehow mentioned generic notions of references, but I cannot find it anymore).

We can indeed think of extracting the collection of reference locations from `cobol_typeck` into a dedicated analysis (possibly in `cobol_lsp` if that's only relevant there). Based on the current types in `Cobol_data` and `Cobol_unit` that may boil down to visiting specialized `resolved_qualname` and `resolved_name` nodes.

EDIT: maybe in a separate PR though.

GBertholon:
`Cobol_unit.Qual.MAP` is used for instance in `Typeck_output.qualrefmap`

GBertholon:
I added a commit that renames `Cobol_unit.Qualmap` into `Cobol_unit.Resolver_map`. I guess this is the last confusing part we can easily solve without more heavy refactoring.

src/lsp/cobol_unit/unit_resolver_map.ml
```diff
@@ -49,7 +49,12 @@ module TYPES = struct
 end
 include TYPES
 
-type +'a t = 'a qualmap
+type +'a t = 'a resolver_map
+
+let () =
+  Printexc.register_printer (function
+    | Ambiguous (lazy qns) -> Some (Pretty.to_string "Ambiguous reference found: %a" (NEL.pp Unit_qual.pp) qns)
+    | _ -> None)
```
nberth: Please fit within 80 columns.

nberth (approved):
Fine!

src/lsp/cobol_unit/unit_resolver_map.ml
GBertholon: I really need a tool that does that for me...

Issue #577, opened 2026-04-17
(1) Creation of the project


pauline-lam:
The goal of this task is to create the basic code infrastructure around the interpreter: an OCaml program that receives one or several COBOL modules in arguments, other arguments to be passed to the COBOL modules, and then, calls functions to prepare the code for execution and calls other functions to perform the interpretation. 

All these functions are just TODO statements, except parsing the code and generating the AST, using existing code from SuperBOL.
Gather a set of small COBOL programs that will be used as first tests, using Crowbar to run them. 
Create a CI/CD to run the tests automatically at each commit, and distribute binaries for several targets _(Linux, Windows, OSX)_.

nberth:
Random notes/tasks to get started:
- [x] Define a (separate?) CI workflow where `libcob5-dev` in installed;
- [x] Start writing OCaml bindings against `libcob5`;
- [x] Define a `tramabol` binary that parses and type-check the COBOL source code using the existing parser;
- [x] Write a small series of tests.

PR #588, opened 2026-04-29
Tentative `ctypes`-based bindings with `libcob`

nberth:
Draft bindings that mostly rely on `ctypes`: bindings are declared using OCaml code, and a `dune` extension generates appropriate C stubs. Incorrect declarations lead to stubs that do not compile properly against `libcob`, which leads to a compile-time error.

Completes the first 2 tasks of #577.


nberth:
Did not convince some people (sadly).

PR #589, opened 2026-04-29
Tentative `camlidl`-based bindings with `libcob`

nberth:
Draft `camlidl`-generated bindings. Overall I think these are harder to get right than the `ctypes`-based in #588: mostly due due to more subtle memory management declarations.


nberth:
Closing as not the solution we're going for.

PR #591, opened 2026-04-30
Another take on tab handling

### LINKED ISSUE #68: Improve accounting of columns in the presence of tabulation characters
At the moment, any tabulation character counts as one towards the right margin in fixed format code. Better accounting is needed to properly detect the portions of text that exceed the right margin (and to take `tab-width` configuration into account).

emilienlemaire:
Should close #530


src/lsp/cobol_preproc/src_lexing.ml
```diff
@@ -357,6 +359,20 @@ let flush_continued ?(force = false) state = match state.continued with
...
+  let byte_col = start_pos.pos_cnum - start_pos.pos_bol in
+  let visual_col = byte_col + 1 - state.current_cpos_shift in    (* 1-indexed *)
+  let tab_len = ((visual_col - 1) mod tab_stop) + 1 - tab_stop in
+  let state =
+    { state with current_cpos_shift = state.current_cpos_shift + tab_len }
+  in
+  if visual_col <= 6 && Option.is_some sna then
+    (Option.value ~default:k sna) state lexbuf
+  else
+    k (flush_continued state) lexbuf
```
nberth: I don't think we should reset info about continued lines here so the call to `flush_continued` is probably to avoid here.

src/lsp/sql_parser/lexer.mll
```diff
@@ -122,8 +122,6 @@ rule token = parse
       { END_EXEC }
   | "end-exec"
       { END_EXEC }
-  | "END-EXEC."
-      { END_EXEC }
```
nberth, line 126: That's indeed a weird entry… but maybe that should be left aside and just marked with a CHECKME comment.

src/lsp/cobol_common/srcloc.ml
```diff
@@ -364,46 +364,80 @@ let pp_file_loc ppf ((file, pos1, pos2): raw_loc) =
...
+      Buffer.add_char buf '^'
+    else if !vcol <= col2 then begin
+      (* col2 is past end of line: fill gap then place ^ *)
+      for v = !vcol to col2 - 1 do
+        Buffer.add_char buf (if v > col1 then '^' else ' ')
+      done;
+      Buffer.add_char buf '^'
+    end;
+    Buffer.contents buf
+  in
```
nberth: For now I'd just remove the tweaks on carets (or just treat tabs as single spaces, which should render "just fine").
That contradicts one of my previous comments in another PR: here the focus is on getting the computation of columns right during the lexing phase; and the rendering could be addressed in a second step if really needed (that's not necessary for the LSP server itself, so not a priority). Fixing one issue at a time simplifies testing.

nberth (commented):
That's great work! Exactly the way to go I think.

src/lsp/sql_parser/lexer.mll
emilienlemaire: This one broke some tests and was easy to change directly

src/lsp/cobol_common/srcloc.ml
emilienlemaire: I know, but for testing purpose this is useful to have a printer that handles tabs well

nberth: If that is for testing purposes, do you think one can add a general tab-rewriting code in the test utilities instead of the tested tool itself?

src/lsp/cobol_preproc/src_overlay.ml
```diff
@@ -202,7 +202,17 @@ let restart ?at ctx =
...
+      if Limit.equal next_left right then
+        (* Self-loop: adjacent tokens share the same adjusted pos_cnum (produced
+           by tab expansion or UTF-8 multi-byte chars).  The next token's
+           right_of entry is still valid — keep it.  Only clear from that
+           token's right limit onwards. *)
+        (match HLnks.find_opt ctx.right_of right with
+         | Some (_, next_right) -> clear_right_gap next_right
+         | None -> ())
+      else
+        clear_right_of next_left
```
nberth: `pos_cnum` in lexical locations *should* always be actual offsets in bytes. I think if an issue manifests itself here that may indicate an improper construction of said locaitons.

src/lsp/cobol_preproc/src_lexing.ml
```diff
@@ -357,6 +359,20 @@ let flush_continued ?(force = false) state = match state.continued with
          stage to account for quotes in comment paragraphs. *)
       emit (AlphanumPrefix { knd; qte; str } &@ loc) (reset_cont state)
 
+let tab ?sna ~k ({ config = { tab_stop; _}; _} as state) lexbuf =
+  let _, start_pos, _ = lexeme_info lexbuf in
+  let byte_col = start_pos.pos_cnum - start_pos.pos_bol in
+  let visual_col = byte_col + 1 - state.current_cpos_shift in    (* 1-indexed *)
```
nberth: I think there is a function somewhere in this file that gives the proper column number.

src/lsp/cobol_preproc/src_lexer.mll
```diff
@@ -257,7 +257,7 @@ rule fixed_line state
       }
```
nberth, line 257: Note: here we use a fixed `sna` pattern. However, it may be worth also creating a dedicated (recursive) `fixed_sna_continuation` rule that recognizes the proper amount of spaces and tabs before switching to `fixed_indicator`, with the help of `Src_lexing.sna_tab` and maybe `Src_lexing.sna_space` (or `Src_lexing.sna_blank ~character` that tests for tab or space)… possibly leaving this fixed `sna` pattern here in case it acts as a shortcut/hot path for performances of the lexer).

Be wary this is a `shortest` lexer rule (contrary to most other rules).

src/lsp/cobol_preproc/src_lexing.ml
```diff
@@ -357,6 +357,15 @@ let flush_continued ?(force = false) state = match state.continued with
          stage to account for quotes in comment paragraphs. *)
       emit (AlphanumPrefix { knd; qte; str } &@ loc) (reset_cont state)
 
+let tab ?sna ~k state lexbuf =
+  let _, start_pos, _ = lexeme_info lexbuf in
+  let byte_col = start_pos.pos_cnum - start_pos.pos_bol in
```
nberth: Note: the actual column number is given by `pos_column: state -> Lexing.position -> int`.

src/lsp/cobol_preproc/src_lexer.mll
```diff
@@ -176,17 +176,17 @@
...
 let newline = '\r'* '\n'
 let nnl = _ # ['\r' '\n']                             (* anything but newline *)
-let sna = nnl nnl nnl nnl nnl nnl              (* 6 chars; TODO: exclude tabs *)
-let spaces = ([' ' '\t']*)
-let    blank        = [' ' '\009' '\r']
-let nonblank        = nnl # blank
-let    blanks       =(blank+ | '\t')
-let    blank_area_A = blank blank blank blanks | '\t'
+let nnl_no_tabs = nnl # [ '\t' ]
+let sna = nnl_no_tabs nnl_no_tabs nnl_no_tabs nnl_no_tabs nnl_no_tabs nnl_no_tabs              (* 6 chars; TODO: exclude tabs *)
```
emilienlemaire: This should be removed now

src/lsp/cobol_preproc/src_lexer.mll
```diff
@@ -251,28 +251,34 @@ let mf_cdir_word =
 
 rule fixed_line state
   = shortest
-  | sna                                                            (* nominal *)
+  | '\t'
       {
-        fixed_indicator (Src_lexing.sna state lexbuf) lexbuf
+        Src_lexing.tab ~sna:fixed_indicator ~k:fixed_nominal_line state lexbuf
```
nberth: From a first glance: this may become `sna_tab` so the optional `sna` argument given to `Src_lexing.tab` may be dropped; this may remove one of the extraneous hard-coded `6`.

src/lsp/cobol_preproc/src_lexer.mll
```diff
@@ -251,28 +251,34 @@ let mf_cdir_word =
...
       }
-  | '\t'
+  | nnl_no_tabs
       {
-        fixed_nominal_line (Src_lexing.flush_continued state) lexbuf
+        Src_lexing.sna_blank ~k_continue:fixed_sna_continuation
+                             ~k_done:fixed_indicator
+                             6 state lexbuf
```
nberth: If supported by lexer rules, that `6` could become `~remaining_cols:6` for better readability.

src/lsp/cobol_preproc/src_lexing.ml
```diff
@@ -43,6 +43,7 @@ type 'k state =
     ignored: lexloc list;               (** lexical locations of ignored text *)
     cdir_seen: bool;
     current_cpos_shift: int;
+    tab_col_shift: int;        (** shift in columns induced by tabs *)
```
nberth: Isn't there a way to avoid using two counters? I think `current_cpos_shift` is enough for the accounting, but maybe I'm wrong on that…

nberth: (yeah… actually not sure at all…)

src/lsp/cobol_preproc/src_lexing.ml
```diff
@@ -357,6 +360,70 @@ let flush_continued ?(force = false) state = match state.continued with
          stage to account for quotes in comment paragraphs. *)
       emit (AlphanumPrefix { knd; qte; str } &@ loc) (reset_cont state)
 
+(** Fixed tab stop positions (1-indexed columns).  Tabs advance to the next
+    listed column; once all stops are exhausted, the last interval (difference
+    between the final two entries) repeats indefinitely. *)
+let tab_stops = [7; 8; 16]
```
nberth: Maybe we should stick to first tests with `[8; 16]` to match `cobc`'s handling.

src/lsp/cobol_preproc/src_lexing.ml
emilienlemaire: I don't think we could, they are used for opposite reason, `current_cpos_shift` is for multi bytes chars that takes one column, and `tab_col_shift` is for a one byte char that takes multiple columns. This would mean having negative values in `current_cpos_shift`, and even if I think we might get away with merging both, having them separate makes more sense to me

nberth: Considering this is kind of a performance-sensitive part of the whole parser, I think it makes sense to give it a go (no worries for the negative shift).

emilienlemaire: Ok i'll try on separate commit so it is easily reversible

src/lsp/cobol_preproc/src_lexing.ml
```diff
@@ -129,12 +130,16 @@ let change_source_format ({ config; _ } as state) sf =
   else Error ()
 
 let pos_column { current_cpos_shift; _ } Lexing.{ pos_bol; pos_cnum; _ } =
-  pos_cnum - pos_bol + 1 - current_cpos_shift            (* count cols from 1 *)
+  pos_cnum - pos_bol + 1 - current_cpos_shift  (* count cols from 1 *)
 
 let adjust_postion { current_cpos_shift; _ } pos =
```
emilienlemaire: @nberth Here when we have tabs in current_cpos_shift, we loose some infos about the utf8 shift. It does not seem to upset any tests, but this is a critical point to check.

nberth: We have to reason about it this way: `current_cpos_shift` holds the number we need to subtract from the  byte offset in the byte stream to obtain the actual column number (when further subtracting `pos.pos_bol`): an *n*-byte utf-8 character increases this shift by *n*-1, whereas a 1 byte character that counts as *n* decreases this shift by *n*-1.

This is exactly what the code below does if we merge `current_cpos_shift` and `tab_col_shitf` appropriately.

test/cobol_parsing/tokens.ml
```diff
@@ -146,3 +146,46 @@ let%expect_test "token-locations-with-missing-program-id" =
...
+       PROGRAM-ID. prog.
+       DATA DIVISION.
+       WORKING-STORAGE SECTION.
+       01 WS-RES PIC XXX.
+       01 WS-ORIG PIC X VALUE 8.
+       PROCEDURE DIVISION.
+       MAIN-PROC.
+       \t\tCOMPUTE WS-RES = WS-ORIG * 12.
+         STOP RUN.
+";
```
nberth, line 164: I think you could add much more tabulations in here.

test/cobol_preprocessing/dune
```diff
@@ -10,7 +10,8 @@
           compiler_directives
           alphanums_with_utf8
```
nberth, line 11: It would make sense to add some tests in `alphanums_with_utf8.ml` with tabs in various places.

test/lsp/lsp_document_symbol.ml
```diff
@@ -500,3 +500,5 @@ let%expect_test "interface-symbol" =
             LINKAGE SECTION(Function)	15:4 -> 16:65
               01 outFactory(Variable)	16:4 -> 16:65
           PROCEDURE DIVISION(Function)	17:4 -> 17:44 |}]
+
+
```
nberth: Please remove these extraneous lines (to reduce the diff).

test/testsuite/sql/sql_exec/sql_test06.cbl
```diff
@@ -142,4 +142,4 @@
            EXEC SQL CONNECT RESET END-EXEC.
 
        100-EXIT. 
-             STOP RUN.
\ No newline at end of file
+             STOP RUN.
```
nberth: Please restore this file for the sake of a minimalist diff.

test/lsp/lsp_testing.ml
```diff
@@ -108,13 +108,61 @@ let make_lsp_project ?toml () =
...
+                    Bytes.set ub (j + 1) '\t'
+                end src_content;
+                "----  " ^ Bytes.to_string ub
+          in
+          (last_tab_content, s' :: acc)
+        else
+          (None, s :: acc)
+      end (None, []) filtered
+    in
+    List.rev lines |>
```
nberth: Could you extract this additional post-processing in a dedicated toplevel function above: so it may be have a clean name, maybe become optional in an easy way, and to avoid unnecessary growing this function.

src/lsp/cobol_preproc/src_lexer.mll
```diff
@@ -251,28 +250,34 @@ let mf_cdir_word =
...
       }
-  | (nnl* newline)                                  (* blank line (too short) *)
+  | (nnl* newline) (* NOTE: remove? other cases are already shorter *) (* blank line (too short) *)
```
nberth: I think this is now equivalent to ` | newline ` (and the one below, to `| eof`. Not extraneous, but simplifiable (the calls to `Src_lexing.sna` can be removed).

Aside, in my opinion `nnl_no_tabs` is not clearer and may even be unclear with the "s". They should be replaced with just `nnl # ['\t']`.

nberth: After this simplification `fixed_line` resembles a lot `fixed_sna_continuation`, isn't it?
So `fixed_line` could almost become
```ocaml
 | epsilon { fixed_sna_continuation 7 state lexbuf }
```
(BTW I'm not sure to like this 7 constant here. Ideally that should be queried via the format — but we can live with it I guess).

lefessan:
https://github.com/OCamlPro/gnucobol/pull/292/

src/lsp/cobol_preproc/src_lexing.ml
```diff
@@ -357,6 +365,71 @@ let flush_continued ?(force = false) state = match state.continued with
          stage to account for quotes in comment paragraphs. *)
       emit (AlphanumPrefix { knd; qte; str } &@ loc) (reset_cont state)
 
+(** Fixed tab stop positions (1-indexed columns). Tabs advance to the next
+    listed column; once all stops are exhausted, the last interval repeats
+    indefinitely. *)
+let tab_stops = [8; 16]
```
nberth: Maybe now align this with the integers now accepted in GnuCOBOL?

test/lsp/lsp_testing.ml
```diff
@@ -95,27 +95,82 @@ let projdir_marker = "__rootdir__"
...
+    end (None, []) lines
+  in
+  List.rev lines
+
 let make_lsp_project ?toml () =
   let projdir = init_temp_project ?toml () in
   let projdir_regexp = Str.(regexp @@ quote projdir) in
   let temp_dir_name = Filename.get_temp_dir_name () in
-  let end_with_postproc expected_output_string =
+  let end_with_postproc ?(expand_carrets=false) expected_output_string =
```
nberth: I think "caret" typically takes a single "r" (not sure…).

test/lsp/lsp_testing.ml
```diff
@@ -95,27 +95,82 @@ let projdir_marker = "__rootdir__"
...
+        | s when String.trim s = "" -> None             (* ignore json RPC header: *)
+        | s when EzString.starts_with ~prefix:"Content-Length: " s -> None
+        | s -> Some (Str.global_replace projdir_regexp projdir_marker s)
+        end
+    in
+    let lines = if expand_carrets then
+      expand_carret_with_tabs filtered
+    else
+        filtered
+    in
```
nberth, line 171: Please fix the formatting here.

src/lsp/cobol_preproc/src_lexer.mll
```diff
@@ -251,28 +251,21 @@ let mf_cdir_word =
...
-      }
-  | (nnl* newline)                                  (* blank line (too short) *)
-      {
-        Src_lexing.new_line (Src_lexing.sna state lexbuf) lexbuf
-      }
-  | (nnl* eof)           (* blank line (too short), without newline character *)
-      {
-        Src_lexing.(flush @@ eof (Src_lexing.sna state lexbuf) lexbuf)
-      }
+  | epsilon { fixed_sna_continuation 6 state lexbuf (*TODO: Parametrize 6 *)}
```
nberth:
```suggestion
  | epsilon
      {
        fixed_sna_continuation 6 state lexbuf (*TODO: Parametrize 6 *)
      }
```

src/lsp/cobol_preproc/src_lexer.mll
```diff
@@ -302,6 +295,29 @@ and fixed_indicator state
...
         gobble_line (Src_lexing.sna state lexbuf) lexbuf
       }
+and fixed_sna_continuation remaining state
+  = parse
+  | '\t'    (* tab in SNA: check expanded col vs indicator position (col 7) *)
+      {
+        Src_lexing.sna_tab
+          ~k_indicator:fixed_indicator
+          ~k_nominal:fixed_nominal_line
+          state lexbuf
```
nberth:
```suggestion
        Src_lexing.sna_tab state lexbuf
          ~k_indicator:fixed_indicator
          ~k_nominal:fixed_nominal_line
```

src/lsp/cobol_preproc/src_lexer.mll
```diff
@@ -302,6 +295,29 @@ and fixed_indicator state
...
+      }
+  | nnl # [ '\t' ]                                             (* SNA space/char *)
+      {
+        Src_lexing.sna_blank ~k_continue:fixed_sna_continuation
+                             ~k_done:fixed_indicator
+                             remaining state lexbuf
```
nberth:
```suggestion
        Src_lexing.sna_blank remaining state lexbuf
          ~k_continue:fixed_sna_continuation
          ~k_done:fixed_indicator
```

src/lsp/cobol_preproc/src_lexing.ml
```diff
@@ -129,12 +130,17 @@ let change_source_format ({ config; _ } as state) sf =
...
+
+let adjust_position ?(base_cpos_shift = 0) { current_cpos_shift; _ } pos =
+  (* Strips only the UTF-8 overhead: (current - base) counts the extra bytes
+     added within the token, leaving the tab contribution (negative, in base)
+     out of the position adjustment. *)
+  let utf8_shift = current_cpos_shift - min 0 base_cpos_shift in
+  if utf8_shift = 0 || position_encoding_in_bytes then
+    pos
+  else
+    Lexing.{ pos with pos_cnum = pos.pos_cnum - utf8_shift }
```
nberth:
```suggestion
  if utf8_shift = 0 || position_encoding_in_bytes
  then pos
  else Lexing.{ pos with pos_cnum = pos.pos_cnum - utf8_shift }
```

src/lsp/cobol_preproc/src_lexing.ml
```diff
@@ -357,6 +365,77 @@ let flush_continued ?(force = false) state = match state.continued with
...
+  in
+  go 1 tab_stops
+
+(** Computes the 0-indexed visual column of the first character after a tab at
+    [start_pos] and returns it together with the updated state. *)
+let compute_tab_shift state (start_pos: Lexing.position) =
+  let col      = pos_column state start_pos in   (* 1-indexed *)
+  let next_col = next_tab_stop col in
+  next_col - 1,
+  { state with current_cpos_shift = state.current_cpos_shift - (next_col - col - 1) }
```
nberth:
```suggestion
  { state with
    current_cpos_shift = state.current_cpos_shift - (next_col - col - 1) }
```

src/lsp/cobol_preproc/src_lexing.ml
```diff
@@ -357,6 +365,77 @@ let flush_continued ?(force = false) state = match state.continued with
...
+let sna_tab ~k_indicator ~k_nominal state lexbuf =
+  let _, start_pos, end_pos = lexeme_info lexbuf in
+  let next_stop, state = compute_tab_shift state start_pos in
+  let state = ignore_lexloc ~start_pos ~end_pos state in
+  if next_stop > 6 then
+    k_nominal (flush_continued state) lexbuf
+  else
+    k_indicator state lexbuf
+
+let tab ?sna ~k state lexbuf =
```
nberth: I think `?sna` is never given at this point, so this argument can be dropped and the function simplified.

src/lsp/cobol_preproc/src_lexing.ml
```diff
@@ -357,6 +365,77 @@ let flush_continued ?(force = false) state = match state.continued with
...
+  next_col - 1,
+  { state with current_cpos_shift = state.current_cpos_shift - (next_col - col - 1) }
+
+(** Handles a non-tab blank character (space) in the SNA area: ignores it in
+    the produced locations and calls [~k_done state lexbuf] when [remaining]
+    reaches 1 (last SNA column), or [~k_continue (remaining-1) state lexbuf]
+    otherwise. *)
+let sna_blank ~k_continue ~k_done remaining state lexbuf =
+  let _, start_pos, end_pos = lexeme_info lexbuf in
+  let state = ignore_lexloc ~start_pos ~end_pos state in
```
nberth, line 407: This function handles spaces in addition to other non-tab characters. I'd rename it `sna_char` (with adjusted comment).

src/lsp/cobol_preproc/src_lexing.ml
```diff
@@ -357,6 +365,77 @@ let flush_continued ?(force = false) state = match state.continued with
...
+    k_done state lexbuf
+  else
+    k_continue (remaining - 1) state lexbuf
+
+(** Handles a tab character in the SNA area: updates [tab_col_shift] and adds
+    the character to the ignored locations list, then dispatches to
+    [~k_indicator] if the expansion lands at or before the indicator column
+    (0-indexed column 6), or to [~k_nominal] (with [flush_continued] applied)
+    if the tab jumped past the indicator column. *)
+let sna_tab ~k_indicator ~k_nominal state lexbuf =
```
nberth, line 418: For this function, I don't think we need to explicitly ignore the associated lexical location (which is just a blank). So the call to `ignore_lexloc` may be dropped.

test/lsp/lsp_testing.ml
```diff
@@ -95,27 +95,83 @@ let projdir_marker = "__rootdir__"
...
+        | s when String.trim s = "" -> None             (* ignore json RPC header: *)
+        | s when EzString.starts_with ~prefix:"Content-Length: " s -> None
+        | s -> Some (Str.global_replace projdir_regexp projdir_marker s)
+        end
+    in
+    let lines = if expand_carets then
+      expand_caret_with_tabs filtered
+    else
+      filtered
+    in
```
nberth:
```suggestion
    let lines =
      if expand_carets then expand_caret_with_tabs filtered else filtered
    in
```

test/cobol_preprocessing/alphanums_with_utf8.ml
```diff
@@ -57,3 +57,64 @@ let%expect_test "alphanumeric-with-utf8-items" =
...
+     Lines 7-8: tab in SNA, prefix; standard continuation closes it
+     Line 9*:   flush comment *)
+  Prog_preproc.show_text ~source_format:(SF SFFixed) {cobol|
+	 "α"                                                              |
+	 "αβ"                                                             |
+     4* complete literals above; prefix and continuation below
+	 "εζ                                                              |
+     6* Missing continuation
+	 "αβ                                                              |
+     8-                                                             "γδ"|
```
nberth, line 112: Is there really a tab here  at the start of line 8?

nberth (commented):
Still some little details to adjust. I don't think tabulation characters within alnum literals are supported yet; or if they are it does not seem tested. That can be left for later though.

src/lsp/cobol_preproc/src_lexing.ml
nberth: In addition, it may be worth matching ' ' to avoid passing a `" "` (single-space) lexeme to `ignore_lexloc`.

src/lsp/cobol_preproc/src_lexer.mll
```diff
@@ -299,16 +302,15 @@ and fixed_sna_continuation remaining state
   = parse
   | '\t'    (* tab in SNA: check expanded col vs indicator position (col 7) *)
       {
-        Src_lexing.sna_tab
+        Src_lexing.sna_tab state lexbuf
           ~k_indicator:fixed_indicator
           ~k_nominal:fixed_nominal_line
-          state lexbuf
       }
   | nnl # [ '\t' ]                                             (* SNA space/char *)
```
nberth: Right above here you can probably match `' '` with  a rule as below, to avoid the unnecessary `ignore_lexloc` for the space character:
```c
  | ' '
      {
        if remaining <= 1
        then fixed_indicator state lexbuf
        else fixed_sna_continuation (remaining - 1) state lexbuf
      }
  | nnl # ['\t' ' ']
     …
```

nberth (approved):
Great! Only the space optimization missing (and an interactive test poking in a running VSCode). After that I think we can merge 👏 .

src/lsp/cobol_preproc/src_diagnostics.ml
```diff
@@ -90,13 +90,22 @@ type warning =
         loc: srcloc;
         item: unexpected_stuff;
       }
+  | Tab_in_alphanum_literal of
+      {
+        loc: srcloc;
+      }
```
nberth, line 96: Better approach, in case you decide to keep this warning: add a `Tab_character_in_alphanum_literal` to `unexpected_stuff` above, and use `Warn_unexpected` instead ;)

src/lsp/cobol_preproc/src_lexer.mll
```diff
@@ -499,6 +539,10 @@ and maybe_fixed_cdir_line c state (* we just read [c='>'] in indicator column *)
       }
 and fixed_continue_line state
   = parse
+  | '\t'
+    {
+      Src_lexing.tab ~flush:false ~k:fixed_continue_line state lexbuf
```
nberth: I'd go with `~k:(fun state -> fixed_continue_line @@ flush_continued state)` or similar, rather than adding an additional flag to `Src_lexing.tab`.

nberth (approved):
Thanks! Great work!

Only a minor adjustment and you can merge.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@
 - Enforcement of Area A checks only when the source format is `COBOL85` [#555](https://github.com/OCamlPro/superbol-studio-oss/pull/555) [#556](https://github.com/OCamlPro/superbol-studio-oss/pull/556)
 - Support for USAGE `COMP-5` and `COMP-6` [#548](https://github.com/OCamlPro/superbol-studio-oss/pull/548) [#606](https://github.com/OCamlPro/superbol-studio-oss/pull/606)
 - Warnings on unsupported usage, where the reported data item size may be be wrong [#570](https://github.com/OCamlPro/superbol-studio-oss/pull/570)
+- Support tabs in source file [#587](https://github.com/OCamlPro/superbol-studio-oss/pull/587)
```
GitMensch: 587 seems to be unrelated
only question: do we have the documentation ready _how_ those are handled / to be configured?

nberth: Ah yes this is 591 indeed.

nberth: At the moment configuration for this aspect is via the `tab-width` setting in a GnuCOBOL configuration file (that is embedded in the VSIX, or manually configured).  User-setting that is not documented yet (it should, probably along with its effect on `cobc` itself).

emilienlemaire:
Ok, I've updated the `tab` function, and edited the CHANGELOG.md file, we should be good now

src/lsp/cobol_preproc/src_lexing.ml
```diff
@@ -402,10 +402,10 @@ let sna_tab ~k_indicator ~k_nominal state lexbuf =
   else
     k_indicator state lexbuf
 
-let tab ?(flush = true) ~k state lexbuf =
+let tab ~k state lexbuf =
   let _, start_pos, _ = lexeme_info lexbuf in
   let _, state = compute_tab_shift state start_pos in
-  k (if flush then flush_continued state else state) lexbuf
+  k state lexbuf
```
nberth: Apologies I though `flush` was true by default… now I think this flag makes sense; it should just be called `before_indicator` instead of `flush` (and be false by default).

CHANGELOG.md
```diff
@@ -15,6 +15,7 @@
 - Amount of references reported in codelens, that wrongly included declarations [#640](https://github.com/OCamlPro/superbol-studio-oss/pull/640)
 - Wrong codelens appearing on `COPY`s [#642](https://github.com/OCamlPro/superbol-studio-oss/pull/642)
 - Grammar rules for `VALUE OF` clause [#629](https://github.com/OCamlPro/superbol-studio-oss/pull/629) (fix for [Issue #625](https://github.com/OCamlPro/superbol-studio-oss/issues/625))
+- Support tabs in source file [#591](https://github.com/OCamlPro/superbol-studio-oss/pull/591)
```
nberth: You can also put that entry at the top instead of bottom of the list.

src/lsp/cobol_preproc/src_lexing.ml
```diff
@@ -402,10 +402,10 @@ let sna_tab ~k_indicator ~k_nominal state lexbuf =
   else
     k_indicator state lexbuf
 
-let tab ~k state lexbuf =
+let tab ?(before_indicator = true) ~k state lexbuf =
```
nberth, line 405: `before_indicator` should default to `false`, and prevent flush when it holds instead.

PR #592, opened 2026-04-30
Remove deprecated Cobol_data.OLD and Cobol_typeck.OLD

GBertholon:
Also move the function name_of_compilation_unit from hidden module Typeck_units to exposed Cobol_ptree.Compilation_group since it can help migrating away from the OLD interface.


src/lsp/cobol_data/package.toml
```diff
@@ -57,7 +57,6 @@ cobol_common = "version"
 cobol_config = "version"
 cobol_ptree = "version"
 zarith = ">=1"
-zarith_stubs_js = ">=0.17"
```
nberth, line 60: Please restore `zarith_stubs_js`: it is necessary for proper compilation of the LSP server into a JS, and it has no effect on native code. (I think there was a comment about that somewhere, but it may have been inadvertently removed by an oldish version of `drom` at some point…).

nberth (approved):
Many thanks for this simplification! Please merge after addressing my only other comment.

src/lsp/cobol_data/package.toml
GBertholon: Ok I am adding a comment back here

PR #594, opened 2026-05-04
Tentative custom-stubs-generator-based bindings with `libcob`

nberth:
This is mostly an import of code by  @lefessan , with:
- adaptations to work on the version of GnuCOBOL4 that's installed as part of the CI;
- simplified management of linking flags for executables that depend on `ezlibcob`…

Note tests of the bindings are currently disabled on macos due to a segfault that still needs to be investigated.


opam/osx/ezlibcob-osx.opam
nberth, line 1: Since `ezlibbob` is unlikely to cross-compile without serious resistance, I'd be in favor of removing `ezlibcob-osx.opam` and `ezlibcob-windows.opam` (via targetted file deletions in `opam-cross` rule).

This way we may be able to prevent opam from having funny ideas in the wrong circumstanceσ…

opam/ezlibcob_test.opam
nberth, line 1: I'm wondering what this package is for?

test/ezlibcob/dune
```diff
@@ -0,0 +1,24 @@
...
+ (action (with-stdin-from test.input
+         (with-stdout-to test.output
+         (run %{exe:test.exe})))))
+
+; Note: add `export HAVE_LIBCOB5 := false` to `Makefile.config`
+; to disable this test and build of its dependencies.
+(rule
+ (alias runtest)
+ (enabled_if (and %{env:HAVE_LIBCOB5=true}               ; <- /!\ enabled by default
+                  (<> %{ocaml-config:system} macosx)))   ; <- Temporarily disabled on macos
```
nberth: Did you try to see what happens if we test on macosx?

nberth (commented):
I did not go into much details yet, but that looks ok for now.

(Posting now so I don't have pending comments lying around for too long).

opam/ezlibcob_test.opam
ddeclerck: No idea ; it was there in your initial commit 😅

nberth: Ouch… we can probably remove it then ;)

opam/osx/ezlibcob-osx.opam
nberth: @ddeclerck Is comment above moot? (i.e does cross-compilation of `ezlibcob` work?)

ddeclerck: Haven't tried yet. I belive it could work...

test/ezlibcob/dune
ddeclerck: As you may have noticed : I removed this with the latest commits ; works fine.

.github/workflows/windows-msys2.yml
```diff
@@ -76,7 +78,8 @@ jobs:
           mkdir import/gnucobol4/_build && cd import/gnucobol4/_build
           # Note: we add the path to git-bash so that we can use the diff
           # command it provides, to workaround a diff bug under the MSYS2 CI
-          PATH="$PATH:/c/Program Files/Git/usr/bin" ../configure $CFGOPT CFLAGS=-std=c17
+          sed -i 's/-Qunused-arguments//g' ../configure
```
nberth, line 81: I did not see this hack before? Is it necessary for a new version of one of the tools?

src/h2mlstubs/error.ml
```diff
@@ -0,0 +1,56 @@
...
+    sym_kind : string ;
+    sym_name : string ;
+    sym_implem : bool ;
+    sym_pos : GoblintCil.location ;
+  }
+
+let errors = Hashtbl.create 111
+
+let mk_error sym fun_name line ty =
+  let ty = GoblintCil.(Pretty.sprint ~width:999999 (Cil.d_type () ty)) in
```
nberth, line 24: Not really important but… may a width of, say 80 or 120, make more sense?

src/h2mlstubs/generate.ml
```diff
@@ -0,0 +1,646 @@
...
+
+  Printf.bprintf ml_buf "  external of_int : int -> t = \"%%identity\"\n";
+  Printf.bprintf ml_buf "  external to_int : t -> int = \"%%identity\"\n";
+  Printf.bprintf ml_buf "  external of_s16 : S16.t -> t = \"%%identity\"\n";
+  Printf.bprintf ml_buf "  external to_s16 : t -> S16.t = \"%%identity\"\n";
+  Printf.bprintf ml_buf "  external of_u16 : U16.t -> t = \"%%identity\"\n";
+  Printf.bprintf ml_buf "  external to_u16 : t -> U16.t = \"%%identity\"\n";
+  Printf.bprintf ml_buf "  external of_s32 : S32.t -> t = \"%%identity\"\n";
+  Printf.bprintf ml_buf "  external to_s32 : t -> S32.t = \"%%identity\"\n";
+  Printf.bprintf ml_buf "  external of_u32 : U32.t -> t = \"%%identity\"\n";
```
nberth, line 569: A little brittle indeed…

src/tramabol/ezlibcob/v1_header.ml
```diff
@@ -0,0 +1,209 @@
...
+  type k
+  type t = k cint
+  external of_int : int -> t = "%identity"
+  external to_int : t -> int = "%identity"
+  external of_char : char -> t = "%identity"
+  external to_char : t -> char = "%identity"
+  let zero = of_int 0
+  let one = of_int 1
+  let inc i = of_int (to_int i + 1)
+  let dec i = of_int (to_int i - 1)
```
nberth: Shouldn't those be named `succ` and `pred` instead, to follow OCaml's standard names?

src/tramabol/ezlibcob/package.toml
```diff
@@ -0,0 +1,174 @@
...
+  (flags (:standard (:include cflags) (:include libcob_cflags)))
+  (include_dirs (include libcob_incdir)))
+(c_library_flags :standard (:include libcob_libs))
+"""
+dune-trailer = """
+
+(executable
+ (name genstubs)
+ (libraries pretty ez_file camlp-streams goblint-cil)
+ (modules error parser types synthetize generate genstubs))
```
nberth: I think this `genstubs` executable is clearly generic enough to deserve being put in a separate package on itself.  Maybe with a name that starts with `c2` (or `h2` if it only reads header files)… like `h2mlstubs` maybe.

src/tramabol/ezlibcob/parser.ml
```diff
@@ -0,0 +1,81 @@
...
+  GoblintCil.initCIL ()
+
+let preprocess_c_file include_files c_file c_file_pp =
+  let null = if Sys.win32 then "NUL" else "/dev/null" in
+  let c_null_in = open_in null in
+  let c_null_out = open_out null in
+  let null_in = Unix.descr_of_in_channel c_null_in in
+  let null_out = Unix.descr_of_out_channel c_null_out in
+  let args =
+    [ "gcc"; "-E";
```
nberth, line 24: Shouldn't a call to `cpp` be enough?

src/h2mlstubs/parser.ml
```diff
@@ -0,0 +1,81 @@
...
+        true
+    | _ ->
+        false
+  in
+  GoblintCil.RmUnused.removeUnused ~isRoot:is_root ci_file;
+  ci_file
+
+let parse_c_file c_file =
+  let c_file_pp = Filename.temp_file (c_file ^ ".") ".pp" in
+  let res = preprocess_c_file [ (* include paths *) ] c_file c_file_pp in
```
nberth, line 88: Maybe remove `c_file_pp` in case of exception?

src/tramabol/ezlibcob/stubs.h
nberth, line 1: May this file miss a header?

nberth (commented):
Thanks for all this extra work on the PR (that I cannot approve as I opened it 🙃 ).

The only "important" comment is about the extraction of the stubs generator, if that's feasible 😉 .

.github/workflows/windows-msys2.yml
ddeclerck: This hack is necessary for the MSYS2/Clang CI. In fact, under any of the MSYS2 environement, opam insists on installing an OCaml-compiler that uses gcc (even if we set the env to clang). However, when GnuCOBOL is compiled with Clang, it adds a few Clang-specific options, which are passed to gcc through the OCaml compiler when linking with libcob. Since gcc does not understand these options, compilation fails. Hence the hack...

src/h2mlstubs/error.ml
ddeclerck: We use the string representation of the type as a key in an hash table (see below). If we put a too low limit, the string will contain an ellipsis (...) for long types. But in practice we are unlikely to encounter very long types.

src/h2mlstubs/generate.ml
ddeclerck: I left this as "TODO" ;)

src/tramabol/ezlibcob/v1_header.ml
ddeclerck: Sure indeed.

src/tramabol/ezlibcob/package.toml
ddeclerck: Yes, but genstubs also relies on some generic "utility" code to be included in the generated C and ML files - right now this code is mixed with handwritten code that is specific to libcob in the header/trailer files. It would have to be moved to dedicated files.

src/tramabol/ezlibcob/parser.ml
ddeclerck: Quite possibly indeed.

src/tramabol/ezlibcob/stubs.h
ddeclerck: Could you be more specific ?

.github/workflows/windows-msys2.yml
nberth: :upside_down_face:

src/tramabol/ezlibcob/package.toml
nberth: Those additional header/trailer files can be left to the user of the executable I guess.

src/tramabol/ezlibcob/stubs.h
nberth: I meant a copyright/license header.

opam/osx/ezlibcob-osx.opam
ddeclerck: But we don't really have time to make it to work.

nberth: If you want to postpone, could you alter the `opam-cross` Makefile rule so these specific `-osx.opam` and `-windows.opam` files do not get created by `make opam-cross`?

src/h2mlstubs/parser.ml
ddeclerck: Actually better to remove it in all cases ; it is no longer needed once parsed.

src/tramabol/ezlibcob/package.toml
ddeclerck: Done. I split them into "utility" code that sits with `h2mlstubs` and libcob-specific code that stays in `ezlibcob`.

src/h2mlstubs/package.toml
```diff
@@ -0,0 +1,99 @@
...
+(rule
+ (targets common_c.ml)
+ (deps common/common.c)
+ (action
+  (with-stdout-to common_c.ml
+   (progn
+    (echo \"let contents = {|\")
+    (cat common/common.c)
+    (echo \"|}\")))))
+"""
```
nberth, line 99: I think we already depend on `ocaml-crunch` via another lib… but that fine as it is now ;)

src/h2mlstubs/parser.ml
```diff
@@ -0,0 +1,94 @@
...
+  in
+  let status_opt =
+    try
+      let pid =
+        Unix.create_process_env "gcc" (Array.of_list args)
+          (Unix.environment ()) null_in null_out null_out
+      in
+      let _pid, status = Unix.waitpid [ (* WNOHANG *) ] pid in
+      Some (status)
+    with _ ->
```
nberth, line 47: I think such catch-all constructs are a bit dangerous (in particular things may break unexpectedly when `Sys.catch_break` is on and we receive a `SIGINT`).

But again, fine with that.

nberth (commented):
Great! many thanks for your adjustments!

src/h2mlstubs/package.toml
ddeclerck: Ah yes indeed, hadn't noticed.

src/h2mlstubs/parser.ml
ddeclerck: Oh, I did not know about `Sys.catch_break`.

ddeclerck (approved):

GitMensch:
so... what does this magic do?

PR #595, opened 2026-05-05
Minor NodeJS binding fix and update

ddeclerck:


src/vendor/vscode-ocaml-platform/src-bindings/node/node.ml
```diff
@@ -54,7 +54,7 @@ module Process = struct
 
     val pid : int [@@js.global "process.pid"]
 
-    val kill : int -> ?signal:string -> unit -> unit [@@js.call "process.kill"]]
+    val kill : int -> ?signal:string -> unit -> unit [@@js.global "process.kill"]]
```
nberth, line 57: Nice catch!

nberth (approved):
Thanks!

src/vendor/vscode-ocaml-platform/src-bindings/node/node.ml
ddeclerck: (actually I'm the one who introduced this error in the first place ;) )

PR #597, opened 2026-05-06
Add the ability to specify a list of copybooks in `Testing_utils`

GBertholon:


nberth (commented):
That's a good start. I suggest to build upon this solution to pass an optional `?copybook:(string * string)` argument instead of `?platform` arguments; this would make things even clearer when reading the tests (no "platform" involved).

What would be even better (but optional) would be to allow passing a list of copybooks (empty by default, of course) instead of a single one.

GBertholon:
OK, I changed the interface to take a copybook list argument

nberth (approved):
Great! Many thanks!

PR #598, opened 2026-05-07
New debugger extension in OCaml

ddeclerck:
This replaces the former debugging extension with an OCaml version.


GBertholon:
A change in the `superbol.debugger.gdbPath` setting is not taken into account until the extension is restarted. I think this is a regression from last version.

GitMensch:
Interesting! How far is this extension? Does it replace the existing features already or are parts missing?
What about portability to non GNU/Linux environments?

Would the DAP allow to use the new debugging backend also on emacs/vim?

Would you consider adding support for cobcd processed files (= commonly no C sources but the "symbols" in a big string + binaries having the COBOL source line references already)?
Note: I'd generally suggest to parse the C files ONLY for mapping the COBOL variable names to the C names - everything else should be available via GDB already (including the COBOL source line mapping since at least 3.2 -> you can set breakpoints to COBOL lines and if you are on a C source line you can just check the previous instruction address until you get a COBOL source line).

Just a reminder: the original fork extension has a test source in which allows to easily check all different data types and let you see its parsing and setting results.

PR #599, opened 2026-05-11
Fix bug in CFG when paragraphs share the same name

GBertholon:
In COBOL if a program contains multiple paragraphs with the same name A, then GO TO A, jumps to the paragraph A in the current section without ambiguity.
The CFG was ignoring this rule and therefore crashed signalling an ambiguity when it occurred.
This PR uses the correct procedure name resolution, and fixes the bug.

### DIFF
--- a/src/lsp/cobol_cfg/cfg_jumps.ml
+++ b/src/lsp/cobol_cfg/cfg_jumps.ml
@@ -41,10 +41,10 @@ module Jumps = Set.Make(struct
       | _ -> to_int j2 - to_int j1
   end)
 
-let full_qn ~cu qn =
-  (Resolver_map.find_binding qn cu.unit_procedure.procedure_blocks.named).full_qn
+let full_qn ~in_section ~cu qn =
+  Procedure.full_qn ?in_section qn cu.unit_procedure
 
-let full_qn' ~cu qn = full_qn ~cu ~&qn
+let full_qn' ~in_section ~cu qn = full_qn ~in_section ~cu ~&qn
 
 
 module JumpsCollector = struct
@@ -61,7 +61,7 @@ module JumpsCollector = struct
                skip_remaining = false;
              }
 
-  let folder ~cu = object (v)
+  let folder ~in_section ~cu = object (v)
     inherit [acc] Visitor.folder
 
     method! fold_goback' _ acc =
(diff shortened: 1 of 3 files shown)


src/lsp/cobol_cfg/cfg_jumps.ml
```diff
@@ -41,10 +41,10 @@ module Jumps = Set.Make(struct
       | _ -> to_int j2 - to_int j1
   end)
 
-let full_qn ~cu qn =
-  (Resolver_map.find_binding qn cu.unit_procedure.procedure_blocks.named).full_qn
+let full_qn ~in_section ~cu qn =
+  Procedure.full_qn ?in_section qn cu.unit_procedure
```
nberth, line 45: `in_section` is a bit ambiguous as a name here. May I suggest using `enclosing_section`, or maybe `within_section`, `from_section`, `origin_section`, or something similar that fits your idea for this option?

nberth: Not mandatory… just the first thing that came to mind when I read it. It may be just a matter of getting used to that name actually…

nberth:
Too bad we don't have tests for that part. What do you estimate would be the effort for adding even a single test for that CFG generation feature, even just by dumping/comparing outputs in graphviz/dot format? In case, in another PR if ever you decided to tackle that ;-)

src/lsp/cobol_cfg/cfg_jumps.ml
GBertholon: I took the name from `Unit_procedure.full_qn`, I also like `enclosing_section` more.

GBertholon:
I guess this is not enough to completely fix the bug...
We still will have the problem if a paragraph in the anonymous section at the beginning of the file refers to an aliased paragraph that appears in the anonymous section.

GBertholon:
The problem stated in my last remark would be fixed by #605

nberth:
Do you think a test that exhibits that issue could be written, for instance via tests for definition lookups in `test/lsp/lsp_definition.ml`?  Maybe by extending the test named "definition-ambiguous-section/paragraphs" in this file?

nberth:
@GBertholon I've pushed a tentative test.  This test does not involve the lookup mechanism used in `cfg_builder`; it's to try and identify the issue with current CFG construction.

src/lsp/cobol_cfg/cfg_jumps.ml
nberth: I've also addressed that.

test/cobol_cfg/ascii_basic.ml
```diff
@@ -0,0 +1,106 @@
...
+    |                |  - +
+    |       A        |    '
+    |                | <- +
+    +----------------+
+      '            ^
+      '            '
+      v            '
+    +-----------+  '
+    |  B IN A   | -+
+    +-----------+ |}];;
```
nberth: @GBertholon I've just checked, this graph was wrong without the fix in this PR.

PR #601, opened 2026-05-19
Drop merlin submodule

nberth:
Gather every grammarware scripts into `src/lsp/grammarware`. Included changes remove symbolic links that may cause issues when trying to compile on some platforms.


src/lsp/grammarware/printer/gen_printer.ml
emilienlemaire, line 1: Shouldn't we also add a license header here?

emilienlemaire: If you don't want to add the usual license header from other `printer/` files, you can just add a comment mentioning the provenance of the file (here `merlin` iirc)

emilienlemaire (approved):
Ok to merge after adding the missing comment / license header

src/lsp/grammarware/printer/gen_printer.ml
nberth: No I forgot to add a header for that file. My bad.

PR #602, opened 2026-05-27
Add an MSYS2 CI for some components of the project

nberth:


.github/workflows/windows-msys2.yml
```diff
@@ -0,0 +1,121 @@
...
+  build:
+    runs-on: windows-latest
+    timeout-minutes: 30
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # - { target: release, sys: mingw64, env: x86_64       }
+          - { target: debug,   sys: mingw64, env: x86_64       }
+          # - { target: release, sys: ucrt64,  env: ucrt-x86_64  }
```
ddeclerck: Do we really want to use the mingw64 runtime (based on the legacy MSVCRT.dll) instead of the new ucrt64 runtime (based on the more modern UCRT.DLL) ?

ddeclerck (approved):

.github/workflows/windows-msys2.yml
nberth: No idea (I don't know how to balance those runtimes' worth).  I can activate that to see what happens in the Ci ;)

ddeclerck: Could be a good idea to activate both debug/mingw64 and release/ucrt64

nberth: Actually do we really mean `release` at this point?

nberth: *need?

PR #605, opened 2026-06-02
Refactoring around the AST for PROCEDURE DIVISION

GBertholon:
- procedure_range does not need to be generic: removing the genericity allows to write visitors that can make action each time a procedure_range is appears in the AST.
- Section names cannot be qualified: replace the qualname with a simple name to reflect that and simplify processing.

I guess we should also simplify the management of the anonymous section at the beginning of a PROCEDURE DIVISION in this PR to remove the annoying difference between paragraphs in this section compared to paragraph in other sections.
This would help fix #599 for real and in general implement more robust analyses.

### DIFF
--- a/src/lsp/cobol_typeck/typeck_procedure.ml
+++ b/src/lsp/cobol_typeck/typeck_procedure.ml
@@ -101,7 +101,7 @@ let procedure_of_compilation_unit ~data_definitions ~refs cu' =
       }
     and section_under_construction =
       {
-        sec_name: Cobol_ptree.procedure_name with_loc;
+        sec_name: string with_loc;
         sec_paragraphs: procedure_paragraph with_loc list;
       }
 
@@ -156,8 +156,8 @@ let procedure_of_compilation_unit ~data_definitions ~refs cu' =
       | None, Some n ->
           { paragraph_name = Some (name n &@<- n);
             paragraph = p } &@<- p
-      | Some { payload = { sec_name = qn; _ }; _ }, Some n ->
-          { paragraph_name = Some (qual n ~&qn &@<- n);
+      | Some { payload = { sec_name; _ }; _ }, Some n ->
+          { paragraph_name = Some (qual n (name sec_name) &@<- n);
             paragraph = p } &@<- p
       | _, None ->
           { paragraph_name = None;
@@ -168,10 +168,10 @@ let procedure_of_compilation_unit ~data_definitions ~refs cu' =
 
     let commit_section acc =
       match acc.current_section with
-      | Some ({ payload = { sec_name = qn; _ }; _ } as section) ->
+      | Some ({ payload = { sec_name; _ }; _ } as section) ->
           let section = section_block section in
           { acc with
-            blocks = Resolver_map.add ~&qn section acc.blocks;
+            blocks = Resolver_map.add (name sec_name) section acc.blocks;
             block_list = section :: acc.block_list;
             current_section = None }
       | None ->
@@ -181,7 +181,7 @@ let procedure_of_compilation_unit ~data_definitions ~refs cu' =
       let acc = commit_section acc in
       let sec_paragraphs = [simple_paragraph s acc] in
       { acc with
-        current_section = Some ({ sec_name = name n &@<- n;
+        current_section = Some ({ sec_name = n;
                                   sec_paragraphs } &@<- s) }
 
     let named_paragraph n p acc =
@@ -195,7 +195,7 @@ let procedure_of_compilation_unit ~data_definitions ~refs cu' =
         | Some s ->
             let loc = Cobol_common.Srcloc.concat ~@s ~@p in
             let sec_paragraphs = p :: ~&s.sec_paragraphs in
-            qual n ~&(~&s.sec_name),
+            qual n (name (~&s.sec_name)),
             acc.block_list,
             Some ({ ~&s with sec_paragraphs } &@ loc)
       in
--- a/src/lsp/cobol_typeck/typeck_procedure.ml
+++ b/src/lsp/cobol_typeck/typeck_procedure.ml
@@ -95,28 +95,22 @@ let procedure_of_compilation_unit ~data_definitions ~refs cu' =
     type acc =
       {
         blocks: procedure_block Resolver_map.t;
-        block_list: procedure_block list;
+        section_list: procedure_section with_loc list;
         current_section: section_under_construction with_loc option;
         args: Cobol_ptree.procedure_using_phrase with_loc option;
       }
     and section_under_construction =
       {
-        sec_name: string with_loc;
+        sec_name: Cobol_ptree.name with_loc option;
         sec_paragraphs: procedure_paragraph with_loc list;
       }
 
     let init =
       {
         blocks = Resolver_map.empty;
-        block_list = [];
+        section_list = [];
         current_section = None;
         args = None;
-      }
-
-    let procedure_blocks acc =
-      {
-        named = acc.blocks;
-        list = List.rev acc.block_list;
       }
 
     let procedure_using acc =
@@ -129,7 +123,8 @@ let procedure_of_compilation_unit ~data_definitions ~refs cu' =
     let procedure acc =
       let procedure_using, references = procedure_using acc in
       {
-        procedure_blocks = procedure_blocks acc;
+        procedure_blocks = acc.blocks;
+        procedure_sections = List.rev acc.section_list;
         procedure_using;
       },
       references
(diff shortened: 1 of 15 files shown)


src/lsp/cobol_typeck/typeck_procedure.ml
```diff
@@ -101,7 +101,7 @@ let procedure_of_compilation_unit ~data_definitions ~refs cu' =
       }
     and section_under_construction =
       {
-        sec_name: Cobol_ptree.procedure_name with_loc;
+        sec_name: string with_loc;
```
nberth: I think this should be `Cobol_ptree.name` instead of a plain `string` (or even a `Cobol_ptree.(name_ term)` — I did not find a good alias for this singleton term type, in case you think about a good one / maybe `Cobol_ptree.given_name`). This is in case we change that type to better deal with case-insensitive comparisons.

src/lsp/cobol_unit/unit_types.ml
```diff
@@ -97,7 +98,11 @@ and arg_passing_style =
 type procedure =
   {
     procedure_using: procedure_using with_loc option; (* PROCEDURE DIVISION USING ... *)
-    procedure_blocks: procedure_block named_n_ordered;
+    procedure_blocks: procedure_block named_n_ordered; 
+    (* FIXME: We should probably replace the ordered part with a regular list of sections 
+       including the implicit anonymous section at begining of the file. 
+       I think we have bugs lurking around linked to this missing section and the 
+       processing irregularity that follows from that. *)
```
nberth: Maybe, if we never need the lookup. Although it should always be noted that (by default) visitors traverse the ordered list which always contains all the elements of a `_ named_n_ordered`; using such visitors is the encouraged way of traversing ASTs here.

There's a comment about that, that is attached to the definition of `named_n_ordered`. We should probably add a `See caveats about {!named_n_ordered}"` doc comment everywhere it is used.

nberth (commented):
I am mostly fine with those changes; one suggestion to address, and maybe a missing comment, before we can merge,

src/lsp/cobol_unit/unit_types.ml
nberth: One more thought: we *could* drop the resolution map and use regular maps to represent the named items of the procedure division. But before that we need to check whether there are constructs that allow one more level of qualification to refer to code blocks (maybe to refer to entries, maybe from sub-programs, etc); this is an aspect I have not investigated deeply.

GBertholon: There is a typo in my comment :( I intended to replace the **unordered** (list) part, like in my next commit in the PR.

Therefore the remark is not about visitor, it is more about the lack of section above every paragraph that can cause problems.

We can discuss whether another kind of map would be more appropriate, but that was out of the scope of this commit.

src/lsp/cobol_typeck/typeck_procedure.ml
GBertholon: Technically `Cobol_ptree.name` and `string` are synonyms, but you are right

src/lsp/cobol_typeck/typeck_procedure.ml
```diff
@@ -95,30 +95,24 @@ let procedure_of_compilation_unit ~data_definitions ~refs cu' =
...
       {
         blocks: procedure_block Resolver_map.t;
-        block_list: procedure_block list;
+        section_list: procedure_section with_loc list;
         current_section: section_under_construction with_loc option;
         args: Cobol_ptree.procedure_using_phrase with_loc option;
       }
     and section_under_construction =
       {
-        sec_name: Cobol_ptree.procedure_name with_loc;
```
nberth, line 104: Could you check whether `Cobol_ptree.procedure_name` is still in use? If yes, I think the good move would be to introduce a new type `Cobol_ptree.procedure_reference` (`= qualname`) and use it instead, and then have `Cobol_ptree.procedure_name = name_ term` (with single constructor `Name` — that would remove some of the additonnal map on Options you needed).

src/lsp/cobol_typeck/typeck_procedure.ml
```diff
@@ -129,84 +123,89 @@ let procedure_of_compilation_unit ~data_definitions ~refs cu' =
...
-            qual n ~&(~&s.sec_name),
-            acc.block_list,
-            Some ({ ~&s with sec_paragraphs } &@ loc)
+      let sec_paragraphs =
+        (* Avoid adding an empty section header paragraph if unneeded, and
+           remove section name from the location of the unnamed paragraph at
+           the begining of section. *)
+        match Cobol_common.Srcloc.concat_locs ~&p.paragraph_sentences with
+        | None -> [] (* when ~&p.paragraph_sentences = [] *)
+        | Some loc -> [{ paragraph_name = None; paragraph = p } &@ loc]
```
nberth, line 185: Although this looks correct, I'm a bit uncomfortable with hiding this kind of processing behind a computation on source locations: even if here it's actually a no-op, it still reads like an actual operation.  To avoid this, even at the cost of slightly larger code (but I hope, clearer — otherwise we can remain at this version), we can add:

```ocaml
let rec concat_srcloc_nel: srcloc Basics.NEL.t -> srcloc = function
  | One loc -> loc
  | loc :: One loc' -> concat loc loc'
  | loc :: loc' :: tl -> concat_srcloc_nel Basics.NEL.(concat loc loc' :: tl)
```

in `cobol_common/srcloc.ml` (that operates on non-empty list of locations only), and here, check for emptiness of `paragraph_sentences` before the operation on source locations.

src/lsp/cobol_unit/unit_types.ml
```diff
@@ -60,22 +61,34 @@ type data_definitions =
...
     data_records: Cobol_data.Types.record list;
   }
 
 (* proc *)
 
 type procedure_paragraph =
   {
-    paragraph_name: Cobol_ptree.procedure_name with_loc option;
+    paragraph_name: Cobol_ptree.procedure_name with_loc option; (* 
+      [None] for the unnamed paragraph at the begining of section.
```
nberth:
```suggestion
      [None] for the unnamed paragraph at the begining of a section.
```

src/lsp/cobol_unit/unit_types.ml
```diff
@@ -60,22 +61,34 @@ type data_definitions =
...
 type procedure_section =
   {
-    section_name: Cobol_ptree.procedure_name with_loc;
-    section_paragraphs: procedure_paragraph with_loc named_n_ordered;
+    section_name: Cobol_ptree.name with_loc option; (*
+      [None] for the unnamed section at the begining of PROCEDURE DIVISION
+      For convenience when collecting references to all procedure names, 
+      [section_name] is visited as a [procedure_name with_loc option]. *)
+    section_paragraphs: procedure_paragraph with_loc named_n_ordered; (*
+      Includes an anonymous empty paragraph if the section is completely empty.
```
nberth:
```suggestion
      Includes an anonymous empty paragraph if the section is empty.
```

Can you explain the rationale behind the introduction of an empty paragraph in empty sections?

src/lsp/cobol_typeck/typeck_procedure.ml
GBertholon: But building a NEL from a list is quite inefficient...
Would you prefer a code like below:
```ocaml
if ~&p.paragraph_sentences = [] then
    []
else
    [{ paragraph_name = None; paragraph = p } &@ 
        Option.get (Cobol_common.Srcloc.concat_locs ~&p.paragraph_sentences)]
 ```
I took the habit to avoid `Option.get` when possible but writing it with an explicit `if` might be more readable and not significantly more inefficient.

Comment (nberth) [reply src/lsp/cobol_typeck/typeck_procedure.ml:185]: Yes possibly (although inefficiency of NEL build can probably be mitigated, there are probably very few paragraphs with >10000 statements).  Still, I am fine with the code you suggest; I also tend to avoid `Option.get`, but here it's fine.

Comment (GBertholon) [reply src/lsp/cobol_typeck/typeck_procedure.ml:185]: Or maybe rewriting the NEL module to have an efficient conversion from list is better ?
If yes, probably in a distinct PR then.

File: src/lsp/cobol_typeck/typeck_procedure.ml (hunk 3 above)
Comment (GBertholon) [reply src/lsp/cobol_typeck/typeck_procedure.ml:104]: Yes, it is used in the `Cobol_ptree` each time a `qualname` is expected to reference a procedure name.
And it corresponds to the associated visitor call `fold_procedure_name`.

File: src/lsp/cobol_typeck/typeck_procedure.ml (hunk 4 above)
Comment (nberth) [reply src/lsp/cobol_typeck/typeck_procedure.ml:185]: I was thinking about that indeed, with a dedicated constructor with list tail… Although this extends the scope of this PR. I'd suggest to use whichever version of the code you prefer, and put a comment (or an issue) with a link to this place in the code and the indications for optimizing list manipulations in NEL.

File: src/lsp/cobol_unit/unit_types.ml (hunk 6 above)
Comment (GBertholon) [reply src/lsp/cobol_unit/unit_types.ml]: The idea is that the target of a GO TO should always be a paragraph to avoid having two different kind of targets.
Therefore I need at least one paragraph per section to be able to jump to a section.
I am not 100% sure this is the right design choice but it can simplify the CFG computation.
The other choice would be to recreate this empty paragraph for analyses that need it and remove this default empty paragraph.

The risk with the current choice is having analyses that need to skip this default empty paragraph.

File: src/lsp/cobol_typeck/typeck_procedure.ml (hunk 3 above)
Comment (nberth) [reply src/lsp/cobol_typeck/typeck_procedure.ml:104]: I think it deserves a rename into `procedure_reference` (with appropriate rename for visitor methods).  Ideally we'd also have `data_reference` when relevant.  Those are changes that would make the code reflect typical terms used in COBOL standards and language reference manuals.

File: src/lsp/cobol_unit/unit_types.ml (hunk 6 above)
Comment (nberth) [reply src/lsp/cobol_unit/unit_types.ml]: I am not sure yet how to balance the two following assumptions' worth:
- assuming every section holds at least a paragraph;
- assuming targets of `GO TO`s may be sections.

At first sight the latter seems to correspond to the COBOL that's written more, (only?) at the cost of an ADT to represent targets of `GOTO`s.  There may be aspects I've not considered; please correct me if there are.

Comment (GBertholon) [reply src/lsp/cobol_unit/unit_types.ml]: If you use an ADT for GO TO targets you then have two different representation for jumping to the anonymous first paragraph of a section. You can decide for a rule such as section target is only used for empty sections or paragraph target is never used for anonymous paragraphs but OCaml type system won't help you avoiding the wrong convention.

I really think that at CFG level I need an anonymous paragraph on empty section. What remains unclear is whether this is a property that we want in general or just for the CFG.

Comment (nberth) [reply src/lsp/cobol_unit/unit_types.ml]: The fact that we can have two representations for this kind of jumps is only an issue for *drawing* CFGs, right?

If that's the case and it's doable, then I'd opt for a small explicit pre-processing pass before CFG computation that introduces such empty paragraphs (only in case unique jump representation is required).  Because doing that too early may prevent a simple linting analysi that could enforce that jumps to the first paragraph of a section are never (or always, not sure what makes the most sense) expressed using the section name.

Comment (GBertholon) [reply src/lsp/cobol_unit/unit_types.ml]: Well not really, a node in the CFG is a paragraph, and at least currently we do not have section nodes (just nodes for anonymous paragraphs at the beginning of section). We may decide to draw section nodes to compress the drawn graph but this is a separate feature on the wish list.

Anyway I don't know what a graph with two levels of nodes means...

Comment (GBertholon) [reply src/lsp/cobol_unit/unit_types.ml]: Your linting example does not seem to work... 

A section can start with an anonymous paragraph or not. Maybe we can rephrase the rule I follow differently to make it sound more natural: a section immediately starting with a paragraph name does not have an anonymous paragraph. All the other sections get an anonymous paragraph.

With the current rule you can write a linter that forbids or forces mentionning the paragraph name if it is the first of the section. You can also provide easily a linter that warns for empty paragraphs and empty sections (an empty section will have an empty anonymous paragraph, and this is the only way of getting one).

Comment (nberth) [reply src/lsp/cobol_unit/unit_types.ml]: My concern is about the treatment of empty sections, for which an artificial empty paragraph is introduced.  There are patterns out there that use empty sections/paragraphs and `PERFORM … THRU …` in specific ways to express some computations.  Proper interpretation of some versions of `PERFORM`s to paragraphs and sections is even the subject of academic papers (and interpretation may even sometimes depend on dialects).  I fear that if we start introducing such an artificial AST node too early we'd run into trouble later to detect those patterns.

File: src/lsp/cobol_unit/unit_types.ml [hunk 7]
```diff
@@ -60,22 +61,34 @@ type data_definitions =
...
   {
-    paragraph_name: Cobol_ptree.procedure_name with_loc option;
+    paragraph_name: Cobol_ptree.procedure_name with_loc option; (* 
+      [None] for the unnamed paragraph at the begining of section.
+      [Some qn] where [qn] is fully qualified otherwise. 
+      Note that [paragraph_name] is different than [~&paragraph.paragraph_name].
+      The latter contains the section name for the optionnal unnamed paragraph 
+      at the begining of the section, and the unqualified paragraph name 
+      otherwise. *)
     paragraph: Cobol_ptree.paragraph with_loc;
```
nberth, line 80: Could you rephrase the above comment?  Can a paragraph name come from a section title?

src/lsp/cobol_unit/unit_types.ml
nberth: > Anyway I don't know what a graph with two levels of nodes means...

For graphical representations that's clusters.  They help a lot in structuring and rendering large graphs and I think as we have this information we should use those for that purpose (even supported by many/most rendering engines).  I don't think it is a good idea to elect to ignore sections already at this point.

src/lsp/cobol_unit/unit_types.ml
GBertholon: Yes, is this new wording better ?

PR #612, opened 2026-06-30
Integrate `autofonce` in source code

nberth:
This removes a submodule, and will help us in dealing with vendored dependencies.


opam/osx/superbol_free_lib-osx.opam
```diff
@@ -57,7 +57,7 @@ depends: [
   "sql_preproc-osx" {= version}
   "lwt-osx" {>= "5"}
   "ez_toml-osx" {= version}
-  "ez_file-osx" {>= "0.3"}
+  "ez_file-osx" {>= "0.4"}
```
emilienlemaire, line 60: Any reason why here we have `0.4` instead of `0.5` like the others?

emilienlemaire (approved):
I suppose that all the `autofonce` code is the same as before so I only read quickly the file with diffs, and skipped the new files

opam/osx/superbol_free_lib-osx.opam
nberth: Just strict adherence to what 0.5 brings over 0.4 (a.k.a `{read,write}_text_file`), that we don't need here (0.4 brought the file path slashifier).

nberth:
> I suppose that all the `autofonce` code is the same as before so I only read quickly the file with diffs, and skipped the new files

Indeed!

PR #615, opened 2026-07-03
Add locations to expressions and conditions and simplify GADTs

GBertholon:
Types expression and condition both now contain locations. The types without location are now respectively called expr and cond.

Remove the GADT and separate visitors for simple_condition vs. complex_condition as they do not really make sense. We might need to reintroduce a GADT for fully expanded conditions later (ie. conditions without Abbrev).

Add an explicit node for parentheses in abbreviated conditions named AbbrevParens. This extra node will be necessary to detect if a syntax is outside a given COBOL dialect.

Simplify the GADT for term by removing the qual_ type argument.


src/lsp/cobol_parser/grammar.mly
```diff
@@ -2801,66 +2803,59 @@ let length_of_expression ==
...
-%inline logop:
- | AND           { LAnd }
- | OR            { LOr }
-
-%inline any_lpar:
- | LPAR              {}
- | LPAR_BEFORE_RELOP {}
+let any_lpar ==
+ | LPAR;              {}
+ | LPAR_BEFORE_RELOP; {}
```
nberth, line 2820: I think those empty may be omitted.

nberth (changes requested):
Thanks for those changes.

My only complaint is that the convention in parse tree type definitions is to use explicit `_ with_loc` types for fields or constructor arguments whenever possible: this way it is easier to find out what is actually source-localized when selecting what data to keep for symbolization of diagnostics.

So I'd avocate for the removal of the `expression` and `condition` types and have their equivalent `_ with_loc` in type definitions.

src/lsp/cobol_parser/grammar.mly
GBertholon: I get a syntax error if I remove the `{}`

nberth: Ah maybe you also need to remove the `;`… otherwise, does not matter much.

GBertholon:
According to your suggestion, I removed the type aliases hiding the with_loc behind longer names.

src/lsp/cobol_ptree/terms_helpers.ml
```diff
@@ -56,7 +57,7 @@ and expand_every_abbrev_condition cond = expand_every_abbrev_cond ~&cond &@<- co
     relation condition encoded by [abbrev_combined_relation] ([= neg,
     relation_condition, logop, flatop]).
 
-    The result is an expression without any abbreviated combined relation
+    The result is an expr with_loc without any abbreviated combined relation
```
nberth: Is this particular substitution intentional?

nberth (commented):
Still fine with those changes ;) except that unfortunately several tests are now failing.

src/lsp/cobol_ptree/terms_helpers.ml
GBertholon: No, it's just me going a bit too fast with my search and replace

GBertholon:
I am working on fixing tests. It seems that we were using the polymorphic (=) to compare expressions and conditions in tests. I think we should ignore locations when comparing AST nodes.

nberth:
> I am working on fixing tests. It seems that we were using the polymorphic (=) to compare expressions and conditions in tests. I think we should ignore locations when comparing AST nodes.

Damn! we wrote dedicated comparison functions for that purpose… there may be places where those functions were not put to use.

GBertholon:
I still have a weird behavioral change on the hints displayed by the recovery system (cf. failing test in the CI).
I am not sure if this is benign (i.e. I should promote the new output) or problematic (i.e. something should be modified in the parser) since at the moment I don't understand very well how the recovery system works.

nberth:
> I still have a weird behavioral change on the hints displayed by the recovery system (cf. failing test in the CI). I am not sure if this is benign (i.e. I should promote the new output) or problematic (i.e. something should be modified in the parser) since at the moment I don't understand very well how the recovery system works.

~~Did you change the grammar?~~
If the hint makes sense it's ok to promote ; it seems to make sense.
The change in behavior may be due to constructs that are no longer amenable to recovery (due to reductions no longer having a dummy recovery action), or become "more expensive" (in terms of the cost model of the underlying synthesis algorithm).

nberth (approved):
Many thanks!

nberth:
Note I think these changes deserve a CHANGELOG entry, possibly something generic like "Improvements to the grammar" under "Added", or more specific "Grammar rules for expressions and conditions" under "Fixed".

PR #617, opened 2026-07-10
Stubgen struct fix

ddeclerck:


src/h2mlstubs/common/common.c
```diff
@@ -256,10 +264,13 @@ CAMLprim value ml_ptr_get(value ptr_v)
 	case CFLOAT: res_v = caml_copy_double(*((float *)data)); break;
 	case CDOUBLE: res_v = caml_copy_double(*((double *)data)); break;
 	case CENUM: res_v = Val_long(*((signed int *)data)); break;
-	case CCOMP:
-		res_v = caml_alloc_small(1, Abstract_tag);
-		Field(res_v, 0) = *((value *)data);
+	case CCOMP: {
+		long size = (long)Field(ptr_v, 2);
+		void *data_copy = NULL;
+		memcpy(data_copy, data, size);
```
nberth, line 270: Not sure to see why a NULL pointer is passed to `memcpy` here…

ddeclerck: Good catch : forgot a malloc here.

src/h2mlstubs/common/common.c
```diff
@@ -476,12 +510,13 @@ CAMLprim value ml_array_of_ptr(value size_v, value ptr_v)
...
 	enum kind_id id = (enum kind_id)Field(sarray_v, 0);
+	long size = (long)Field(sarray_v, 2);
 	void *sdata = (void *)Field(sarray_v, 3);
 	long spos = Long_val(spos_v);
 	void *ddata = (void *)Field(darray_v, 3);
 	long dpos = Long_val(dpos_v);
 	long len = Long_val(len_v);
-	long elem_size = kind_size(id);
+	long elem_size = kind_size(id, size);
 	memcpy(ddata + dpos * elem_size, sdata + spos * elem_size, len * elem_size);
```
nberth: I think the equivalent function in the `Stdlib` allows overlapping areas in the same array.  So I suggest to use `memmove` here instead.

nberth (approved):
Thanks! Ok to merge modulo use of `memmove` in array/blit.

src/h2mlstubs/common/common.c
ddeclerck: Indeed ! Fixed.

PR #621, opened 2026-07-20
[Rebase] New debugger extension in OCaml

Halbaroth:
This PR is an attempt to merge #598.

A few regressions were found in #598, which prevented us from merging it. This PR fixes the following bugs:
1. The output of target is correctly handled by GDB.
2. The `gdbPath` option can be updated without restarting vscode.
3. GDB commands are correctly handled by the vscode debug console. 

To test this PR, run 
```console
make vsix-debug
code --install-extension superbol-vscode... && code 
```

The vscode debug console isn't the default for the target output. You need to edit your `launch.json` must 
contain `"gdbtty": "vscode"` as follows:  
```json 
    "configurations": [
    {
        "name": "SuperBOL: debug (launch)",
        "type": "superbol-gdb",
        "request": "launch",
        "preLaunchTask": "SuperBOL: build (debug)",
        "target": "${file}",
        "arguments": "",
        "cwd": "${workspaceFolder}",
        "gdbtty": "vscode"
    }
    ],
```


Halbaroth:
@ddeclerck I'll remove the flake as soon as the PR is accepted.

ddeclerck (commented):
Regarding the wrapper, note that we aim to support Windows, so a solution has to be found for this platform.

Also note that in the future, the wrapper could be extended to help properly handle external terminals (properly build a pty with forkpty/openpty, instead of resorting to the sleep hack).

Halbaroth:
According to the documentation of `exec-wrapper`: 
> This command is available when debugging locally on most targets, excluding DJGPP, Cygwin, and MS Windows.

:(

ddeclerck:
> According to the documentation of `exec-wrapper`:
> 
> > This command is available when debugging locally on most targets, excluding DJGPP, Cygwin, and MS Windows.
> 
> :(

Then this leaves you only one option: on Windows, use a wrapper-like program that you manage manually ; ie start this wrapper directly, not through gdb. The wrapper should deal with the streams (you could use named pipes for instance), then start the actual program using `CreateProcess` with the `CREATE_SUSPENDED` flag, and finally display the target program pid ; your extension then use that pid to *attach* to the target program (instead of using the "run" command), followed by a `resume` command.

src/vscode/superbol-debugger/sourceMap.ml
```diff
@@ -0,0 +1,560 @@
...
+            (* Match instruction lines between C and COBOL (.c) *)
+            else if Util.matches reProcedure line &&
+                    not (Util.hasGroup 2) then
+              let cobLine = Util.groupInt line 1 in
+              (* Discard last line if it is for the same file&line.
+                 Unlikely, but happens if several statements on same line. *)
+              let lines = Util.filterHd (fun (l : line) ->
+                  l.cobFile <> lacc.cobFile || l.cobLine <> cobLine
+                ) gacc.lines
+              in
```
Halbaroth: @ddeclerck can you explain these lines? If several statements are on the same line, we expect that the debugger stops before the first statement and the line. For instance, if your line is:
```cobol 
...
b+> MOVE 10 TO VAR DISPLAY "Hello".
...
``` 
you expect that VAR is unmodified until you continue the execution of the program. Currently, the break point is between the move statement and the display statement. 

Removing this piece of code gives the right behavior. Can we ignore a line if its line number is already in the map?

ddeclerck: I don't know, it's just a translation of the original code:
[[here](https://github.com/OlegKunitsyn/gnucobol-debug/blob/72c3a17c3ffd9fc65617c2415ffee590ea954500/src/parser.c.ts#L110)](https://github.com/OlegKunitsyn/gnucobol-debug/blob/72c3a17c3ffd9fc65617c2415ffee590ea954500/src/parser.c.ts#L110)
If it works better without it, feel free to remove it.

Halbaroth:
I tested another approach to support redirection of the target outputs on Windows. I thought that I could use a named pipe in the `run` gdb command but redirections are simply ignored here, even if the target is a regular file. I believe that we can merge this PR as soon as I rebase it and the Windows support can be done in a separate PR.?

Halbaroth:
@ddeclerck I rebased the PR on master.

PR #627, opened 2026-08-07
Stubgen : support function pointers through trampolines

ddeclerck:


src/h2mlstubs/generate.ml
```diff
@@ -610,6 +624,153 @@ let gen_enum (enum : enum) =
...
+  else
+    Printf.bprintf c_buf
+      "  res_v = caml_callbackN_exn(cb_%s_closures[slot], %d, args_v);\n"
+      fun_.fun_sig nb_args;
+
+  (* Handle the return value *)
+  begin
+    match fun_.fun_ret with
+    | STVoid (_) ->
+        Printf.bprintf c_buf "  CAMLreturn0;\n"
```
nberth, line 721: Shouldn't a check for exceptions be inserted here as well?

src/h2mlstubs/common/common.c
```diff
@@ -69,6 +90,58 @@
...
+	CAMLparam1(callback_v);
+	f_generic *callback = Callback_val(callback_v);
+	int slot = callback_lookup(callback, cb_table, size);
+	Field(callback_v, 0) = (value)NULL;
+	if (slot >= 0) {
+		caml_remove_generational_global_root(&cb_closures[slot]);
+		cb_closures[slot] = Val_unit;
+	}
+	CAMLreturn(Val_unit);
+}
```
nberth, line 141: I'm wondering whether `inline`-ing `callback_free` and `callback_create` really makes sense…

src/h2mlstubs/synthetize.ml
```diff
@@ -289,8 +370,79 @@ let synthetize_comp loc ci =
...
+              fun_sig = fsig;
+              fun_ret;
+              fun_args;
+            }
+          in
+          Hashtbl.replace env.funs fsig fun_;
+          fun_ :: funs
+  | TVoid (_) | TInt (_) | TFloat (_) | TEnum (_)
+  | TNamed (_) | TBuiltin_va_list (_) ->
+      []
```
nberth, line 445: I though `goblint-cil` exposed a visitor for this kind of traversal… maybe wrongly.

src/h2mlstubs/synthetize.ml
```diff
@@ -289,8 +370,79 @@ let synthetize_comp loc ci =
         fi.fname, synthetize_type ~for_struct:true sym (fi.ftype, fi.fattr)
       ) ci.cfields
   in
-  comp.comp_fields <- fields;
-  comp
+  comp.comp_fields <- fields
+
+
+
+let rec synthetize_fun_types ?(for_fun=false) loc ty =
```
nberth: What does `for_fun` mean here? I think it may deserve a more explicit name.  Is it "for function declaration" as opposed to "for function pointer type"?

src/h2mlstubs/synthetize.ml
```diff
@@ -289,8 +370,79 @@ let synthetize_comp loc ci =
...
+        let funs = synthetize_fun_types loc rt in
+        let funs =
+          List.fold_left (fun funs (_an, ty, _al) ->
+              synthetize_fun_types loc ty @ funs
+            ) funs args
+        in
+        if for_fun then
+          funs
+        else
+          let sym = Error.{
```
nberth, line 416: Why is `Error` opened / why is this type defined there?

src/h2mlstubs/generate.ml
ddeclerck: Yes indeed.

src/h2mlstubs/common/common.c
ddeclerck: Maybe not.

src/h2mlstubs/generate.ml
ddeclerck: Ah, wait no. At least for now, since we just discard the exception. Exceptions should really not escape from callbacks.
Now if you'd like, I could store this exception in some global that could be queried via a function such as `get_last_exception` (but you sometimes don't know when these callbacks will be called and thus when it is appropriate to check for these exceptions).

src/h2mlstubs/synthetize.ml
ddeclerck: This just carries information about the "symbol" we're handling ; this is used by `synthetize_arg` (called below) to report errors.

src/h2mlstubs/synthetize.ml
ddeclerck: Maybe ; though visitors don't immediately come to my mind when I have to do these traversals - too used to good old pattern matching.

src/h2mlstubs/synthetize.ml
ddeclerck: Yup. I'll adjust the name.

src/h2mlstubs/generate.ml
nberth: If I understand correctly, this code is executed on a return from a callback, right?  Any exception raised by such a callback is lost and only prevents an assignment to `res` (below), which is left uninitialized…

Thinking "aloud"… what about having a dedicated exception handler registered via [`Stdlib.Callback.register`](https://ocaml.org/manual/5.5/api/Callback.html). Then the OCaml code could just register a (single) function that should neither return anything or raise any exception, to handle the given exception right here in case `Is_exception_result(res_v)` holds.  The only remaining question would be the value to return from that point… would it make sense to pass a `~result_on_exception` value to `Callback.create`?

src/h2mlstubs/synthetize.ml
nberth: Yes usually, except here you had to implement a flag to avoid infinite recursion, that departs from the usual traversal pattern :wink:

nberth (approved):

PR #628, opened 2026-08-10
Libcob bindings: add checks on integer operations

ddeclerck:


src/h2mlstubs/common/common.ml
```diff
@@ -20,130 +20,194 @@ type !'a comp
...
+end
+
 module S8 = struct
   type k
   type t = k cint
-  external of_int : int -> t = "%identity"
-  external to_int : t -> int = "%identity"
+  include MakeCheckedInteger
+      (BoundedInteger (struct let min = -128 let max = 127 end))
+      (struct type nonrec t = t end)
```
nberth, line 163: Could it be worth adding `min_int` and `max_int` values in those `S8`, `U8`, `S16`… modules?

src/h2mlstubs/common/common.ml
```diff
@@ -20,130 +20,194 @@ type !'a comp
...
+    else Ok (Int64.to_int r)
+  let add a b =
+    let s = Int64.add a b in
+    if Int64.unsigned_compare s a < 0 then
+      Error (IntegerOverflow)
+    else Ok s
+  let sub a b =
+    if Int64.unsigned_compare b a > 0 then
+      Error (IntegerOverflow) else
+      Ok (Int64.sub a b)
```
nberth: Logic seems right; only indentation should be fixed 😅

nberth (approved):
Thanks!

src/h2mlstubs/common/common.ml
ddeclerck: Absolutely. Just done so, and also added a `compare` function.

src/h2mlstubs/common/common.ml
ddeclerck: Fixed.

src/h2mlstubs/common/common.ml
```diff
@@ -20,130 +20,209 @@ type !'a comp
 type !'a cptr
 type !'a carray
 
+type error =
+  | IntegerOverflow
+
+module type CHECKED_ARITH = sig
+  type repr
+  val min : repr
+  val max : repr
```
nberth: `min` and `max` would be functions to take min/max values from two operands I think.  I suggested `min_int` and `max_int` (as in `Int64` module, typically)… maybe `min_val`/`max_val` would also fit if you prefer.

That's not a big deal, but I think following the `Stdlib`'s conventions would help.

ddeclerck: Absolutely ; done !

PR #633, opened 2026-08-13
Extend bindings

engboris:


nberth (approved):
Do you have information about minimal vscode version requirements for these fields to be defined?  Unless those were introduced before 1.64.0.

engboris:
> Do you have information about minimal vscode version requirements for these fields to be defined? Unless those were introduced before 1.64.0.

It should be this: https://code.visualstudio.com/updates/v1_55#_new-extension-property-on-extensioncontext

PR #638, opened 2026-08-18
Parse abbreviated conditions in EVALUATE WHEN clauses

GBertholon:
This commit also removes ambiguous Relation vs Abbrev distinction in cond and unifies SelCond and SelRelation by using the more permissive abbrev_relation_operand in SelCond.

### DIFF
--- a/src/lsp/cobol_ptree/terms_helpers.ml
+++ b/src/lsp/cobol_ptree/terms_helpers.ml
@@ -18,11 +18,21 @@ open Srcloc.INFIX
 let neg_condition ~neg (c: 'r cond with_loc): 'r cond =
   if not neg then ~&c else Not c
 
-let cast_no_rel_cond (c: no_rel cond): 'r cond = 
-    (* OCaml typechecker is not clever enough to deduce that no_rel cond and 
-       'r. 'r cond are equivalent. We use Obj.magic to allow the cast without 
-       rebuilding recursiveley the full condition. *)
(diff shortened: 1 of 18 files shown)


src/lsp/cobol_ptree/terms_helpers.ml
```diff
@@ -15,9 +15,15 @@ open Terms
...
 
-let neg_condition ~neg (c: cond with_loc): cond =
+let neg_condition ~neg (c: 'r cond with_loc): 'r cond =
   if not neg then ~&c else Not c
 
+let cast_no_rel_cond (c: no_rel cond): 'r cond = 
+    (* OCaml typechecker is not clever enough to deduce that no_rel cond and 
+       'r. 'r cond are equivalent. We use Obj.magic to allow the cast without 
+       rebuilding recursiveley the full condition. *)
+    Obj.magic c
```
nberth: I'm wondering whether a "nicer" way to do that is to actually implement the version that does the recursive copy (to show equivalence), and then to bind `cast_no_rel_cond` again with an external `"%identity"`.

nberth (approved):
Great improvements, thanks!

src/lsp/cobol_ptree/terms_helpers.ml
GBertholon: Yes indeed, I have done what you suggested.

PR #634, opened 2026-08-13
Add support for 78-level data items

nberth:
78-level data items define compilation variables.

Included changes bring:
- Checks of VALUE clauses for 78-level items;
- Evaluation of compilation variables in pre-processing directives;
- LSP-related requests about compilation variables (definitions, references, hover, documentation).

As an aside, some internal code refactoring for data literal representation is also included.

Removes limitations mentioned in comments of #161.

PR #642, opened 2026-08-20
Avoid showing codelenses on positions that correspond to COPYs

nberth:
~~Currently based on #634.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Support for 78-level data items [#634](https://github.com/OCamlPro/superbol-studio-oss/pull/634)
 
 ### Fixed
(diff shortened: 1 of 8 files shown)


engboris:
Why do you want to avoid showing codelenses in this case?

nberth:
> Why do you want to avoid showing codelenses in this case?

The change is best illustrated by [this test](https://github.com/nberth/superbol-studio-oss/blob/e18dc4c63307746394128d283870d8db931331d0/test/lsp/lsp_codelens.ml#L217): before, the amount of references for the last (I think) item defined in `lib.cpy` was shown as a codelens on the line with `COPY lib.`.  This was misleading.  Instead we could show some specific info on the whereabouts of the included copybook, or whatever else information we could find useful (but that's for a later PR).

engboris (approved):

PR #644, opened 2026-08-24
Missing dependencies

Ninjapouet:


nberth:
Thanks for these suggestions.  The version constrains you copied mostly apply for `menhirLib` that is used by the packages for which a "tool" dependency was missing.  May a looser constrain (i.e, without the upper-bound) be enough for the tool itself?

Ninjapouet:
Probably but since all the packages are linked together, having the same constraint on the tool and its lib is less confusing I think. It can also simplify the build depending on how the dependencies are set in `menhir`.

nberth (approved):
Thanks!

PR #653, opened 2026-09-14
Add a command to analyze all COBOL programs of the workspace

engboris:
I added a naive command which analyzes the whole workspace for LSP diagnostics (as if we were opening and closing files of the workspace).

It opens a notification with a progress bar. The notification an be hidden and the progression will still we shown on the status bar below.

swappy-20260914-150158


src/lsp/superbol_free_lib/vscode_extension.ml
```diff
@@ -611,6 +611,10 @@ let commands =
       ~title:"Write Project Configuration"
       ~category:"SuperBOL"
     (* ~enablement:"!inDebugMode" *);
+    Manifest.command ()
+      ~command:"superbol.analyze.workspace"
+      ~title:"Analyze All COBOL Files in Workspace"
```
nberth:
```suggestion
      ~title:"Analyze Every COBOL File in the Workspace"
```
Feels more "English" to be ("all" is less frequent I think).

nberth (approved):
Nice, thanks!

nberth (approved):

PR #657, opened 2026-09-15
Improve handling and file in workspace analysis and add reporting

engboris:
1. The strategy for file opening has been improved. Instead of just opening all files naively, files are opened then closed after retrieval of diagnostics;
2. Diagnostics is reported in a `_superbol/analysis-report.md` file so that some information are persistent even in case of interruption. The point of the `.md` is that with the integrated markdown preview in VS Code, we can click on references to jump to it directly (right-click then `Open Preview` or `Ctrl+Shift+V` on the report file).


src/vscode/superbol-vscode-lib/superbol_workspace.ml
```diff
@@ -39,60 +38,194 @@ let find_cobol_files ~token =
...
+  | folder :: _ ->
+      let dir =
+        Node.Path.join [Uri.fsPath (WorkspaceFolder.uri folder); "_superbol"]
+      in
+      let file = Node.Path.join [dir; "analysis-report.md"] in
+      try
+        (* The bindings drop the label: this is a plain, non-recursive mkdir,
+           and it fails if the directory is already there. *)
+        if not (Node.Fs.existsSync dir) then
+          Node.Fs.mkdirSync dir ~recursive:true;
```
nberth: "Non-recursive mkdir" comment does not match the code…

src/vscode/superbol-vscode-lib/superbol_workspace.ml
```diff
@@ -39,60 +38,194 @@ let find_cobol_files ~token =
...
+  | DiagnosticSeverity.Error -> "error"
+  | DiagnosticSeverity.Warning -> "warning"
+  | DiagnosticSeverity.Information -> "note"
+  | DiagnosticSeverity.Hint -> "hint"
+
+(* Markdown link, so that a click jumps to the reported line.  The target is
+   relative to the report, which lies one directory below the workspace. *)
+let report_link ~path ~line ~char =
+  let target = if Node.Path.isAbsolute path then path else "../" ^ path in
+  Printf.sprintf "[%s:%u:%u](<%s#L%u>)" path line char target line
```
nberth, line 157: Warning: the path may not be in the workspace directory where the report is generated…

nberth (approved):
The only comment may not be that hard to solve ;)

PR #665, opened 2026-09-17
Add size and offset on hover for fields

engboris:
When hovering field (either variables of field definitions), the offset and size of the field is displayed. For the offset, it is relative to some top-level field mentioned at hover. For instance, `Offset: 54 bytes in SOME-FIELD`.

Screenshot from 2026-09-17 14-48-56
Screenshot from 2026-09-17 14-49-13

I also renamed the previous `show_hover_text_on_definitions` boolean flag into `show_data_description_on_definitions`.


src/lsp/cobol_lsp/lsp_data_info_printer.ml
```diff
@@ -312,3 +320,35 @@ let pp_data_definition ppf = function
...
+    if not has_issues then
+      let record = enclosing_record ~record_name:(named_record record) qualname in
+      Fmt.pf ppf "%s%a  \n%a" prefix (pp_offset_in record) offset pp_size size
+  in
+  match def with
+  | Data_field { def; record }
+  | Data_condition { field = def; record; _ } ->
+      let x = ~&def in
+      pp ppf ~has_issues:x.field_has_definition_issues ~pp_size ~size:x.field_size
+        ~offset:x.field_offset x.field_qualname record
```
nberth: This may not be right in the case of 88-level/condition items.
To simplify and possibly unify bits of this code, functions from `Cobol_data.Item` may be used.

nberth (approved):
Approved, subject to one minor refactoring suggestion and possibly a fix for 88-level items.

src/lsp/cobol_data/data_item.ml
```diff
@@ -100,7 +100,7 @@ let def_offset: data_definition -> Data_memory.offset = function
 
 let def_has_issues: data_definition -> bool = function
   | Data_field { def; _ } -> ~&def.field_has_definition_issues
-  | Data_renaming _ -> false
+  | Data_renaming { def; _ } -> ~&def.renaming_has_definition_issues
```
nberth, line 103: Ouch 😅

PR #666, opened 2026-09-17
Fixes on the tab handling 👺

emilienlemaire:


nberth (approved):
Nice fix. Thanks!

Please append a reference to this PR in the entry about support for tabulation characters in the CHANGELOG.  Then you can merge.

nberth:
@emilienlemaire I don't understand the rationale for the late changes. `tab-stops` was a better name for the feature, and the computation code should not go in `cobol_config`.  If really needed somewhere else than `cobol_preproc`, maybe that can be fine as `Cobol_common.Tab_stops`.

nberth (changes requested):
For now I think you can revert the two last commits.  Tests might still be worth keeping, after alterations to avoid hitting `cobol_config`.

emilienlemaire:
I'm ok to revert the second to last commit, but without the last one, we would have to put tab width config as a string (`"6,1,8"`) in the config, instead of as an int list (`6,1,8`).

emilienlemaire:
Also in GnuCOBOL, the config name is `tab-width`, hence my choice of name, and I then normalized it over the all project for consistency

nberth:
> Also in GnuCOBOL, the config name is `tab-width`, hence my choice of name, and I then normalized it over the all project for consistency

We will at some point/soon remove most unneeded dependencies to current `cobol_config` (that shall become `gnucobol_config` and only be used to derive more succinct configurations for preproc & parse libraries).  I'm totally fine to keep the fix to the config parser; to do so you can put `Tab_stops` in `cobol_common`.

nberth (approved):
Thanks again. Please squash the last three commits before merging.

